PR #1, opened 2018-10-16
bootstrap the bookstore package

rgbkrk:
`depot` was taken, as was some other good options for this package. Since I also own `bookstore` and that hasn't been updated to the latest versions of jupyter, I'd like to just launch this project as `bookstore` v2.0.0.


setup.py
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+""""
+setup.py
+
+Note: Do a version check for IPython.
+    IPython v6+ no longer supports Python 2.
+    If Python 2, intall ipython 5.x.
```
willingc: install

rgbkrk:
Sweet, thanks! I think that typo probably exists in papermill too, as that's where I copied the package layout from. 🤖

Issue #7, opened 2018-11-09
Create config that is shared across Archiver, Publisher, etc.


rgbkrk:
Since we're going to want to use the same S3 settings across the Jupyter extension for publishing and the archival contents manager, we need to be able to instantiate a config object that can be shared amongst these.

Right now this is encapsulated in `bookstore/bookstore_config.py` as `BookstoreS3Settings` and `Bookstore `. What I need is how we'll rely on these within `BookstoreContentsArchiver` and how we'll set them up in the `jupyter_notebook_config.py` (example in the `README`).

cc @mpacer

mpacer:
In #8 I put together what should be a basic example of how you could set these things up (I also updated the example configs to take that into account). It will need to look slightly different for handlers (since handlers are not themselves `LoggingConfigurable`s we cannot pass them down through `parent=self`). 

Let me know if there's anything more that would benefit from clarification :)

rgbkrk:
This is now done for the archiver, @mpacer and I talked a bit in chat about how we'll handle this for the handlers case (since, as M stated, we can't hook in with `parent=self` within `__init__`).

PR #8, opened 2018-11-12
Config improvements and examples

mpacer:
In response to #7 I thought I'd put together an example of how you could get access to all of this from within a single configurable object.

If this is merged into production we should drop 991118a, as it is not useful code but meant to illustrate what is happening.


rgbkrk:
Closing in favor of #9. Thank you for these alternatives and explanation to help get this through.

PR #11, opened 2018-11-13
set up minio on circle

### LINKED ISSUE #12: Automate Integration Tests
* Set up minio, either in a Docker image or otherwise
* Setup a jupyter notebook config that uses the access keys and endpoints for the minio service
* Write a notebook via the contents API
* Verify the file is written to:
  * Minio
  * The local file system

rgbkrk:
Closes #12.

* [x] Set up minio in a container on CircleCI
* [x] Create a packaged release for testing
* [x] Install packaged release into a virtualenv (full end-to-end integration test)
* [x] Fix bug in handlers registration (this can be moved to another PR if requested)
* [x] Write integration test that launches jupyter with bookstore enabled and tests that the notebook was written to disk and "s3" (minio) properly

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,14 +6,15 @@ version: 2
 jobs:
   build:
     docker:
-      # specify the version you desire here
-      # use `-browsers` prefix for selenium tests, e.g. `3.6.1-browsers`
-      - image: circleci/python:3.6.1
+      - image: circleci/python:3.6.7-node-browsers
+        ports: 9988:9988
 
-      # Specify service dependencies here if necessary
-      # CircleCI maintains a library of pre-built images
-      # documented at https://circleci.com/docs/2.0/circleci-images/
-      # - image: circleci/postgres:9.4
+      - image: minio/minio:RELEASE.2018-11-06T01-01-02Z
+        command: server /data
+        environment:
+          MINIO_ACCESS_KEY: ONLY_ON_CIRCLE
+          MINIO_SECRET_KEY: CAN_WE_DO_THIS
+        ports: 9000:9000
 
     working_directory: ~/repo
 
@@ -23,19 +24,41 @@ jobs:
       # Download and cache dependencies
       - restore_cache:
           keys:
-          - v1-dependencies-{{ checksum "requirements.txt" }}
+          - v2-dependencies-{{ checksum "requirements.txt" }}-{{ checksum "requirements-dev.txt"}}
           # fallback to using the latest cache if no exact match is found
-          - v1-dependencies-
+          - v2-dependencies-
 
       - run:
           name: install dependencies
           command: |
             python3 -m venv venv
             . venv/bin/activate
+            pip install --upgrade pip setuptools wheel
             pip install -r requirements.txt
             pip install -r requirements-dev.txt
 
       - save_cache:
           paths:
             - ./venv
-          key: v1-dependencies-{{ checksum "requirements.txt" }}-{{ checksum "requirements-dev.txt" }}
+          key: v2-dependencies-{{ checksum "requirements.txt" }}-{{ checksum "requirements-dev.txt" }}
+
+      - run:
+          name: package up bookstore
+          command: |
+            . venv/bin/activate
+            # Package up the package
+            python setup.py sdist bdist_wheel
+      - run:
+          name: create virtual environment for packaged release
+          command: |
+            python3 -m venv venv_packaged_integration
+            . venv_packaged_integration/bin/activate
+            pip install --upgrade pip setuptools wheel
+            pip install -U --force-reinstall dist/bookstore*.whl
+      - run:
+          name: integration tests
+          command: |
+            . venv_packaged_integration/bin/activate
+            # Install the dependencies for our integration tester
+            npm i
+            node ci/integration.js
--- a/ci/integration.js
+++ b/ci/integration.js
@@ -1,0 +1,236 @@
+const child_process = require("child_process");
+const { randomBytes } = require("crypto");
+const path = require("path");
+const fs = require("fs");
+
+global.XMLHttpRequest = require("xmlhttprequest").XMLHttpRequest;
+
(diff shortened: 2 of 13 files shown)


.circleci/config.yml
```diff
@@ -34,8 +35,25 @@ jobs:
...
       - save_cache:
           paths:
             - ./venv
-          key: v1-dependencies-{{ checksum "requirements.txt" }}-{{ checksum "requirements-dev.txt" }}
+          key: v2-dependencies-{{ checksum "requirements.txt" }}-{{ checksum "requirements-dev.txt" }}
+
+      - run:
+          name: integration tests
+          command: |
+            python3 -m venv venv2
```
willingc: This is probably a bit nit picky. Can we change `venv2` to `venv_tests`? I keep thinking of Python 2 when I see the name and want to make sure others aren't confused.

ci/jupyter_notebook_config.py
```diff
@@ -0,0 +1,21 @@
...
+# At ~/.jupyter/jupyter_notebook_config.py for user installs
+# At __ for system installs
+c = get_config()
+
+c.NotebookApp.contents_manager_class = BookstoreContentsArchiver
+
+c.BookstoreSettings.workspace_prefix = "ci-workspace"
+
+# If using minio for development
+c.BookstoreSettings.s3_endpoint_url = "http://127.0.0.1:9000"
```
willingc: Do we also need to test https:// ? Probably doesn't matter at this stage, but to keep in mind for later tests.

.circleci/config.yml
rgbkrk: This is a great suggestion, I've even confused myself.

ci/jupyter_notebook_config.py
rgbkrk: We can visit that later. I'm going a little overboard with integration tests here already. 😅

.circleci/config.yml
```diff
@@ -34,8 +35,25 @@ jobs:
             . venv/bin/activate
             pip install -r requirements.txt
```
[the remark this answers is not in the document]
rgbkrk: Ah, perfect, thanks. I'll both upgrade wheel + setuptools as well as update the manifest.

rgbkrk:
So close now. Need to automate creation of the bucket for our use now.

ci/jupyter_notebook_config.py
willingc: Works for me ;-)

willingc:
A little closer... looks like an object isn't being stored perhaps.

rgbkrk:
😬  I was not setting up promises with my minio call properly.

rgbkrk:
It works! Now I can clean this mess up.

.circleci/config.yml
```diff
@@ -6,14 +6,15 @@ version: 2
 jobs:
   build:
     docker:
-      # specify the version you desire here
-      # use `-browsers` prefix for selenium tests, e.g. `3.6.1-browsers`
-      - image: circleci/python:3.6.1
+      - image: circleci/python:3.6.7-node-browsers
+        ports: 9988:9988
```
rgbkrk: This can be removed now.

ci/integration.js
```diff
@@ -0,0 +1,236 @@
+const child_process = require("child_process");
+const { randomBytes } = require("crypto");
+const path = require("path");
+const fs = require("fs");
+
+global.XMLHttpRequest = require("xmlhttprequest").XMLHttpRequest;
+
+const { ajax } = require("rxjs/ajax");
+
+const rxJupyter = require("rx-jupyter");
```
rgbkrk: `rx-jupyter` isn't actually used here, need https://github.com/nteract/nteract/pull/3651 to be merged and shipped.

ci/integration.js
```diff
@@ -0,0 +1,236 @@
...
+
+const _ = require("lodash");
+
+console.log("running bookstore integration tests");
+
+async function genToken(byteLength = 32) {
```
rgbkrk: I like how I went a little extra by using `randomBytes` then just encoded the minio secrets in the config out of what I expected to be a necessity.

ci/integration.js
```diff
@@ -0,0 +1,236 @@
...
+      `--port=${jupyterPort}`,
+      `--ip=127.0.0.1`
+    ],
+    { cwd: __dirname }
+  );
+
+  ////// Refactor me later, streams are a bit messy with async await
+  // Check to see that jupyter is up
+  let jupyterUp = false;
+
```
rgbkrk: I could make this all a bit cleaner with rxjs + spawn-rx. As long as folks are cool with it for now, I'm ok leaving all this since it's an integration test.

ci/integration.js
willingc: Nice!

ci/integration.js
willingc: Let's mention rxjs + spawn-rx in the refactor comments. Just in case another contributor does the refactor

rgbkrk: Ok, will do.

rgbkrk:
Ok, cool. I've refactored this into some separate modules. I probably broke something in the process so let's see how this is.

ci/jupyter.js
```diff
@@ -0,0 +1,101 @@
+const { genToken } = require("./token");
+const { sleep } = require("./sleep");
+
+// "Polyfill" XMLHttpRequest for rxjs' ajax to use
+global.XMLHttpRequest = require("xmlhttprequest").XMLHttpRequest;
+const { ajax } = require("rxjs/ajax");
+
+class JupyterServer {
+  constructor(config = {}) {
+    this.port = this.config.port || 9988;
```
rgbkrk: Note to self when I get back from workout -- this needs to just be config.port

.circleci/config.yml
rgbkrk: Oops, this can be in a new PR.

rgbkrk: Done in #13

PR #20, opened 2018-11-15
Perform S3 Writes on Threads

rgbkrk:
Current plan:

* [x] Offload each write to a thread
* [x] Don't write to S3 when already writing for a path

### DIFF
--- a/bookstore/archive.py
+++ b/bookstore/archive.py
@@ -1,9 +1,21 @@
 import json
-import s3fs
+from asyncio import Lock
+from concurrent.futures import ThreadPoolExecutor
+from typing import Any, Dict, NamedTuple
 
+import s3fs
 from notebook.services.contents.filemanager import FileContentsManager
+from tornado import gen, ioloop
+from tornado.concurrent import run_on_executor
+from tornado.util import TimeoutError
 
 from .bookstore_config import BookstoreSettings
+
+
+class ArchiveRecord(NamedTuple):
+    filepath: str
+    content: str
+    queued_time: float
 
 
 class BookstoreContentsArchiver(FileContentsManager):
@@ -29,6 +41,47 @@ class BookstoreContentsArchiver(FileContentsManager):
             s3_additional_kwargs={},
         )
 
+        self._thread_pool = ThreadPoolExecutor(max_workers=self.settings.max_threads)
+
+        # A lock per path to suppress writing to S3 when currently writing
+        self.path_locks: Dict[str, Lock] = {}
+        # Since locks are not pre-created,
+        self.ability_to_create_path_lock = Lock()
+
+    async def archive(self, record: ArchiveRecord):
+        """Process a storage write, only allowing one write at a time to each path
+        """
+        async with self.ability_to_create_path_lock:
+            lock = self.path_locks.get(record.filepath)
+
+            if lock is None:
+                lock = Lock()
+                self.path_locks[record.filepath] = lock
+
+        # Skip writes when a given path is already locked
+        if lock.locked():
+            self.log.info("Skipping archive of %s", record.filepath)
+            return
+
+        async with lock:
+            try:
+                self.log.info("Processing storage write of %s", record.filepath)
+                full_path = self.s3_path(record.filepath)
+                await self.write_to_s3(full_path, record.content)
+                self.log.info("Done with storage write of %s", record.filepath)
+            except Exception as e:
+                self.log.error(
+                    'Error while archiving file: %s %s',
+                    record.filepath,
+                    e,
+                    exc_info=True,
+                )
+
+    @run_on_executor(executor='_thread_pool')
+    def write_to_s3(self, full_path: str, content: str):
+        with self.fs.open(full_path, mode="wb") as f:
+            f.write(content.encode("utf-8"))
+
     @property
     def delimiter(self):
         """It's a slash! Normally this could be configurable. This leaves room for that later,
@@ -38,7 +91,9 @@ class BookstoreContentsArchiver(FileContentsManager):
     @property
     def full_prefix(self):
         """Full prefix: bucket + workspace prefix"""
-        return self.delimiter.join([self.settings.s3_bucket, self.settings.workspace_prefix])
+        return self.delimiter.join(
+            [self.settings.s3_bucket, self.settings.workspace_prefix]
+        )
 
     def s3_path(self, path):
         """compute the s3 path based on the bucket, prefix, and the path to the notebook"""
@@ -50,12 +105,16 @@ class BookstoreContentsArchiver(FileContentsManager):
         if model["type"] != "notebook":
             return
 
-        # TODO: store the hash of the notebook to not write on every save
-        notebook_contents = json.dumps(model["content"])
+        content = json.dumps(model["content"])
 
-        full_path = self.s3_path(path)
+        loop = ioloop.IOLoop.current()
 
-        # write to S3
-        # TODO: Write to S3 asynchronously to not block other server operations
-        with self.fs.open(full_path, mode="wb") as f:
-            f.write(notebook_contents.encode("utf-8"))
+        # Offload the archival to be scheduled to work on the event loop
+        loop.spawn_callback(
+            self.archive,
+            ArchiveRecord(
+                content=content,
+                filepath=path,
+                queued_time=ioloop.IOLoop.current().time(),
+            ),
+        )
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,6 +36,15 @@ jobs:
             pip install -r requirements.txt
             pip install -r requirements-dev.txt
 
+      - run:
+          name: check types
+          command: |
+            . venv/bin/activate
+            pip install mypy==0.641
+            # NOTE: We can add in stubs for modules we're using
+            # BONUS: Tornado 6 will include types
+            mypy bookstore --ignore-missing-imports
+
       - save_cache:
           paths:
             - ./venv
--- a/bookstore/bookstore_config.py
+++ b/bookstore/bookstore_config.py
@@ -1,5 +1,4 @@
-from traitlets import Any, Bool, Dict, Instance, List, TraitError, Type, Unicode, validate, default
-
+from traitlets import Int, Unicode
 from traitlets.config import LoggingConfigurable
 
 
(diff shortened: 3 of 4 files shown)


bookstore/archive.py
```diff
@@ -29,6 +35,45 @@ def __init__(self, *args, **kwargs):
...
+                print("Processing storage write of %s" % path)
+                notebook_contents = json.dumps(model["content"])
+
+                full_path = self.s3_path(path)
+
+                # write to S3
+                with self.fs.open(full_path, mode="wb") as f:
+                    f.write(notebook_contents.encode("utf-8"))
+
+                await gen.sleep(0.01)
```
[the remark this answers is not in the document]
rgbkrk: This isn't actually a blocking call here, as it makes a tornado coroutine.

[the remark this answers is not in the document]
rgbkrk: When L69 runs, it blocks. I put this sleep in just to see if coroutines played well in this non `async` function.

bookstore/_version.py
```diff
@@ -52,34 +51,38 @@ class NotThisMethod(Exception):
     """Exception raised if a method is not valid for the current scenario."""
 
 
-LONG_VERSION_PY = {}
-HANDLERS = {}
+LONG_VERSION_PY: dict = {}
+HANDLERS: dict = {}
```
rgbkrk: Introduced annotations here to make `mypy` happy (yup, kicking off type annotations).

bookstore/_version.py
```diff
@@ -515,6 +547,10 @@ def get_versions():
...
-    return {"version": "0+unknown", "full-revisionid": None,
-            "dirty": None,
-            "error": "unable to compute version", "date": None}
+    return {
+        "version": "0+unknown",
+        "full-revisionid": None,
+        "dirty": None,
+        "error": "unable to compute version",
+        "date": None,
+    }
```
rgbkrk: Rest of the changes here are all `black`. I can run `black` over all the source in another PR.

bookstore/archive.py
```diff
@@ -52,12 +107,16 @@ def run_pre_save_hook(self, model, path, **kwargs):
         if model["type"] != "notebook":
             return
 
-        # TODO: store the hash of the notebook to not write on every save
-        notebook_contents = json.dumps(model["content"])
+        content = json.dumps(model["content"])
 
-        full_path = self.s3_path(path)
+        # run_pre_save_hook isn't async, though this should be awaited. I need to look into how to integrate with
```
slackpad: Sill noodling on this. When we are trying to put things into the queue to tee up a save we can call [`put_nowait()`](https://www.tornadoweb.org/en/stable/queues.html#tornado.queues.Queue.put_nowait) which will just tell us no, which is ok because we are too busy to try saving and we are only attempting to archive. The auto save from Jupyter will try again later if the user has the notebook open, right (is that always enabled)? If that's not always true to eventually clean up when the queue is full then I think we will need something like their `fetching` set and a little fancier arrangement.

rgbkrk: Yeah the `put_nowait` makes sense in this case, then we'd just catch the `QueueFull` exception. We'd still let this function terminate so the regular save happens.

rgbkrk:
I've switched this all over to callbacks that get registered and made use of tornado's lock primitive to not write when a prior is writing.

It just dawned on me I don't need to use the tornado lock though, as the locks could be tracked in a `set` of filepaths.

bookstore/_version.py
rgbkrk: Went ahead and did this in #21

rgbkrk:
This is ready for a solid review.

bookstore/archive.py
```diff
@@ -5,6 +5,22 @@
 
 from .bookstore_config import BookstoreSettings
 
+from concurrent.futures import ThreadPoolExecutor
+
+from tornado import gen, ioloop
+from tornado.locks import Lock
+from tornado.queues import Queue
+from tornado.concurrent import run_on_executor
+from tornado.util import TimeoutError
```
charsmith: Small nitpick, but shouldn't we sort the imports?

bookstore/archive.py
```diff
@@ -29,6 +45,46 @@ def __init__(self, *args, **kwargs):
             s3_additional_kwargs={},
         )
 
+        self._thread_pool = ThreadPoolExecutor(max_workers=30)
```
charsmith: Worth documenting the magic number or better yet making it configurable?

bookstore/archive.py
```diff
@@ -29,6 +45,46 @@ def __init__(self, *args, **kwargs):
...
+
+    async def archive(self, record: ArchiveRecord):
+        """Process a storage write, only allowing one write at a time to each path
+        """
+        lock = self.locks.get(record.filepath)
+
+        try:
+            if lock is None:
+                lock = Lock()
+                self.locks[record.filepath] = lock
```
charsmith: I admit that I haven't used tornado or their locks, but this pattern seems broken.  If two threads try to lock a new filepath, they could both end up at line 59 at the same time.  Both will assign the lock at the same time (well sort of, last one would set the lock, but both threads would still have their own), and then you will have a race condition on write.  It isn't necessarily that bad, but I think that if we were being correct there should be a global lock that is required to create a new lock:

```
 if lock is None:
     async with self.locks.get('global'):
          lock = self.locks.get(record.filepath)
          if lock is None:
              lock = Lock()
              self.locks[record.filepath] = lock
```

Although I would probably just init that lock at the object level as that is a little too magical/coupled for my taste.

bookstore/archive.py
```diff
@@ -29,6 +45,46 @@ def __init__(self, *args, **kwargs):
...
+
+            # Try to acquire the lock, and if it fails don't bother archiving
+            # as we should finish a previous write
+            await lock.acquire(1)
```
charsmith: I would recommend putting this whole lock in a with block.  If anything would be written before the try block below, an exception would cause the lock to never be released.

rgbkrk: Yeah, I could refactor to do that. I was trying to separate `tornado`'s `TimeoutError` (desired behavior when there's an in-progress write) out here with the steps that shouldn't fail.

rgbkrk:
Thank you so much for the review @charsmith!

bookstore/archive.py
rgbkrk: Configurable is probably the right approach, as I stuck this in here just to have something to work with for the moment.

rgbkrk:
I've made some messy little changes to the [`tornado` `lock`s](https://www.tornadoweb.org/en/stable/locks.html) in some uncommitted changes, I'll come back to this one in the morning.

bookstore/archive.py
rgbkrk: Probably worth noting -- we're using threads here for one specific action that would block the event loop, which is writing to S3. We can't use the `Lock`s or `Queue`s in any of that threaded code.

From [tornado docs on locks](https://www.tornadoweb.org/en/stable/locks.html):

> Note that these primitives are not actually thread-safe and cannot be used in place of those from the standard library–they are meant to coordinate Tornado coroutines in a single-threaded app, not to protect shared objects in a multithreaded app.

bookstore/archive.py
```diff
@@ -29,6 +41,46 @@ def __init__(self, *args, **kwargs):
...
         )
 
+        self._thread_pool = ThreadPoolExecutor(max_workers=30)
+
+        self.path_locks: Dict[str, Lock] = {}
+
+    async def archive(self, record: ArchiveRecord):
+        """Process a storage write, only allowing one write at a time to each path
+        """
+        lock = self.path_locks.get(record.filepath)
```
slackpad: I think this still has the potential for the race that Charles was worried about because two routines might want to create a new lock for the same path. To close that hole we'd need a lock, initialized when this object is made, that protects `self.path_locks` while we add a new path in there.

bookstore/archive.py
```diff
@@ -29,6 +41,46 @@ def __init__(self, *args, **kwargs):
...
+
+        except TimeoutError as e:
+            self.log.info("Another archival in progress for %s", record.filepath)
+            return
+
+        # At this point we have the lock and can write to S3
+        try:
+            self.log.info("Processing storage write of %s", record.filepath)
+            full_path = self.s3_path(record.filepath)
+            await self.write_to_s3(full_path, record.content)
```
slackpad: Although, thinking about this more, since we have an `await` in here, won't this block the `archive()` call as if we just talked to S3 here directly? I was thinking we'd do a `queue.put()` in here with a low timeout, and have a separate routine that pulls from the queue and does the IO to S3.

bookstore/archive.py
rgbkrk: I definitely agree, haven't gotten back to this one yet.

bookstore/archive.py
rgbkrk: Tornado is on a single threaded event loop so this will allow any other archive calls, notebook server ops, etc. to go ahead. The `write_to_s3` is wrapped in a decorator that makes it return a Future that will resolve when the blocking code we've moved to threads is finished up.

rgbkrk: I don't follow.

rgbkrk:
I've switched this over to `asyncio`'s `Lock` as it has a `locked()` method that can be used synchronously to check if a lock is already acquired (so that in our case we can just return). Additionally, I've put in the `ability_to_create_path_lock` that @charsmith suggested.

Bit of an aside -- it might be worth it for us to check out https://github.com/aio-libs/aiobotocore at some point for S3, though it looks like they're just getting started.

bookstore/archive.py
slackpad: I'm still new to Tornado and Python's coroutine model but it looks like the `write_to_s3()` activity will occur in another thread but this thread that calls that will block here because of the `await` so essentially this is like a synchronous call to `write_to_s3()` here.

rgbkrk: It will block in the sense that code won't run below, that's true. It frees up other CPU bound work to run though, new API calls, serving assets, as well as any kernel communication processing.

The code on this branch is a little bit different, but I'll attempt to describe it.

`archive` submits the `write_to_s3` call effectively like this:

```python
future = executor.submit(write_to_s3, full_path, record.content)
await future # our own execution stops, but doesn't block other work on the ioloop
```

bookstore/archive.py
rgbkrk: This is now configurable.

bookstore/archive.py
slackpad: Got it - thanks for the explanation - I think I've got my head around this model now :-)

bookstore/archive.py
```diff
@@ -29,6 +41,47 @@ def __init__(self, *args, **kwargs):
...
+            if lock is None:
+                lock = Lock()
+                self.path_locks[record.filepath] = lock
+
+        # Skip writes when a given path is already locked
+        if lock.locked():
+            self.log.info("Skipping archive of %s", record.filepath)
+            return
+
+        async with lock:
```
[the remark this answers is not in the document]
rgbkrk: That's correct.

rgbkrk:
Thanks for all the reviews folks!

PR #33, opened 2019-01-24
refactors archive functionality to use aiobotocore & asyncio

mpacer:
It is remarkable how much nicer & cleaner this API is.

client.put_object seems to expect/allow json content encoded as a `str` (rather than requiring us to reëncode of json content back to `bytes`). I'm not sure if that's a good thing or a bad thing, but if it's a bad thing we need to probably update our tests to capture the case where that causes a problem.

### DIFF
--- a/bookstore/archive.py
+++ b/bookstore/archive.py
@@ -1,17 +1,15 @@
 import json
-from asyncio import Lock
-from concurrent.futures import ThreadPoolExecutor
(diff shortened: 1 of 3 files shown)


rgbkrk:
To your comment about string encoding -- yeah that seems like something we should test for, for the sake of stability.

mpacer:
You're so fast!

PR #37, opened 2019-01-28
Publish integration

mpacer:
This adds functionality for publishing Notebooks to ci-published and then to compare what is stored there against our in memory copy.

Also adds an integration test for that functionality.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,6 +71,7 @@ jobs:
           command: |
             . venv_packaged_integration/bin/activate
             # Install the dependencies for our integration tester
+            jupyter serverextension enable --py bookstore
             npm i
             node ci/integration.js
 
--- a/.gitignore
+++ b/.gitignore
@@ -114,3 +114,5 @@ binder/*run*.ipynb
 jupyter_notebook_config.py
 jupyter_config.py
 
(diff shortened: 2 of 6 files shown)


mpacer:
Ok so it's definitely the publishNotebook xhr request that is off, so I'm going to try to iterate on that in python for a bit.

mpacer:
So the problem was that we hadn't enabled the serverextension on CI, and it doesn't currently happening automatically (currently working on a PR to fix this).

I added it to the README as well as to the CI and now it works gloriously.

The reason it caused a problem is that the request was expecting JSON back, but because the endpoint wasn't valid, the server was returning a 404 response with an html page.

It'd be nice if we could have a better error than 
```
unhandledRejection SyntaxError: Unexpected token < in JSON at position 0
    at JSON.parse (<anonymous>)
```
but there may be other reasons why that could occur and it would be overkill to build something special purpose for this case.

But I figure that this is a good place to document that if anyone else runs into `Unexpected token < in JSON at position 0`, a good first guess is that it's getting an html page instead of a JSON object… and that might be because you're hitting an endpoint that the server doesn't think exists.

rgbkrk:
Thanks for taking this through the paces @mpacer!

PR #38, opened 2019-01-29
Automatically enable serverextension if one is not using an editable install

mpacer:
add data_files to setup.py so it automatically enables serverextension

also: cleans up setup.py os.path imports

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,7 +71,6 @@ jobs:
           command: |
             . venv_packaged_integration/bin/activate
             # Install the dependencies for our integration tester
-            jupyter serverextension enable --py bookstore
             npm i
             node ci/integration.js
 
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -71,7 +71,6 @@ jobs:
           command: |
             . venv_packaged_integration/bin/activate
             # Install the dependencies for our integration tester
-            jupyter serverextension enable --py bookstore
```
rgbkrk: Magic! 🎩

jupyter_config/jupyter_notebook_config.d/bookstore.json
```diff
@@ -0,0 +1,7 @@
+{
+  "NotebookApp": {
+    "nbserver_extensions": {
+      "bookstore": true
+    }
+  }
+}
```
rgbkrk: 🙈

PR #39, opened 2019-01-29
Tests

mpacer:
prototype branch to see if this hack from https://github.com/pypa/pipenv/issues/2421#issuecomment-400330048 solves the problem


willingc:
Do you want to close this @mpacer now that #40 is merged? A new PR with any JavaScript tweaks or docs would be helpful 👍

PR #40, opened 2019-01-29
Remove versioneer

willingc:


willingc:
As of commit 6e1c6e3, all the Python tests and mypy stuff passes. Recommending merging this and then doing a second PR to handle the JS packaging for Circle CI and integration tests.

willingc:
Adding back @mpacer's changes in an earlier PR solved the last puzzle piece. Yay M.

bookstore/_version.py
```diff
@@ -1,543 +1,17 @@
...
+    "dev",  # dev or nothing
+)
+
+# pep 440 version: no dot before beta/rc, but before .dev
+# 0.1.0rc1
+# 0.1.0a1
+# 0.1.0b1.dev
+# 0.1.0.dev
+
+__version__ = ".".join(map(str, version_info[:3])) + ".".join(version_info[3:])
```
rgbkrk: Works for me. I'm assuming the next step is something like bumpversion?

willingc: Min and I just have that as part of the checklist and manually bump. That said, Python doesn't release as frequently as JS so we can add that if needed.

PR #44, opened 2019-01-29
add simple validate_bookstore function & tests (not traitlets validate)

mpacer:
@rgbkrkr asked to have a way to test for whether bookstore is properly configured over in the nteract_on_jupyter handler.

I felt like the business logic for that should live in the bookstore repo… seemed like the best way to do that would be to expose a validation function.

Since it requires that a collection of values are set appropriately, I figured it was going to be more straightforward to write a quick function rather than define custom validation functions for each one of the traitlets in question.

In the long run, we might want to go in the traitlets direction… but, for now, this seemed like a good way to delay validation and expose it to another library.

### DIFF
--- a/bookstore/bookstore_config.py
+++ b/bookstore/bookstore_config.py
@@ -39,6 +39,13 @@ class BookstoreSettings(LoggingConfigurable):
 
 
 def validate_bookstore(settings: BookstoreSettings):
+    """Test for establishing that bookstore has been appropriately set up.
+    
+    Parameters
+    ----------
+    settings: bookstore.bookstore_config.BookstoreSettings
+        The instantiated Settings object to be validated.
+    """
     valid_settings = [settings.workspace_prefix != "",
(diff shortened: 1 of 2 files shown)


bookstore/bookstore_config.py
```diff
@@ -36,3 +36,12 @@ class BookstoreSettings(LoggingConfigurable):
         16,
         help="Maximum number of threads for the threadpool allocated for S3 read/writes",
     ).tag(config=True)
+
+
+def validate_bookstore(settings: BookstoreSettings):
```
willingc: Let's go ahead and add a simple docstring for the autodocs. Examples are in the papermill docs PR that is soon to merge. numpy format.

mpacer: Great idea!

rgbkrk:
Sweet, does this mean we just pluck the bookstore settings object off the app config in nteract on jupyter then use this validation function?

PR #46, opened 2019-01-30
Add detail to archive docstrings and Sphinx docs

willingc:
- edits `archive.py` docstrings
- proposes changing `ability_to_create_path_lock` to `path_lock_ready`
- Autodoc all source files for Sphinx 

[Rendered docs](https://test-bookstore.readthedocs.io/en/archive-update/index.html)

### DIFF
--- a/bookstore/archive.py
+++ b/bookstore/archive.py
@@ -77,8 +77,8 @@ class BookstoreContentsArchiver(FileContentsManager):
         allowed to a path if a valid path_lock is held and the path is not
         locked by another process.
 
-        Attribute
-        ---------
+        Parameters
+        ----------
         record : ArchiveRecord
             A notebook and where it should be written to storage
         """
@@ -121,8 +121,8 @@ class BookstoreContentsArchiver(FileContentsManager):
         When the event loop is available for execution of the request, the
         storage of the notebook will be done and the write to storage occurs.
 
-        Arguments
-        ---------
+        Parameters
+        ----------
         model : str
             The type of file
         path : str
--- a/bookstore/archive.py
+++ b/bookstore/archive.py
@@ -1,30 +1,37 @@
+"""Archival of notebooks"""
 import json
-import aiobotocore
-
 from asyncio import Lock
 from typing import Dict, NamedTuple
 
+import aiobotocore
 from notebook.services.contents.filemanager import FileContentsManager
 from tornado import ioloop
 
 from .bookstore_config import BookstoreSettings
-
 from .s3_paths import s3_key, s3_display_path
 
 
 class ArchiveRecord(NamedTuple):
+    """Representation of archived content"""
     filepath: str
     content: str
     queued_time: float
 
 
 class BookstoreContentsArchiver(FileContentsManager):
-    """
-    Archives notebooks to S3 on save
-    """
+    """Archives notebooks to S3 on save by UI or parameterized workflow
 
+    Bookstore settings and Jupyter application settings are shared
+    A session is created for the current event loop
+    To write to a particular path on S3, acquire a path_lock.
+    The lock avoids multiple writes contending to write at the same time.
+
+    After acquiring the lock, `archive` authenticates using the storage service's
+    credentials. If allowed, the notebook is written to storage (i.e. S3).
+    """
     def __init__(self, *args, **kwargs):
         super(FileContentsManager, self).__init__(*args, **kwargs)
+
         # opt ourselves into being part of the Jupyter App that should have Bookstore Settings applied
         self.settings = BookstoreSettings(parent=self)
 
@@ -34,7 +41,14 @@ class BookstoreContentsArchiver(FileContentsManager):
             )
         )
 
-        self.session = aiobotocore.get_session()
+        # TODO: create a better check for credentials - keys and config or it will fail
+        try:
+            # create a session object from the current event loop
+            self.session = aiobotocore.get_session()
+        except Exception as e:
+            self.log.error(
+                "{} : Check credentials and configuration".format(e)
+            )
 
         # A lock per path to suppress writing to S3 when currently writing
         self.path_locks: Dict[str, Lock] = {}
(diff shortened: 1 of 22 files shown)


bookstore/archive.py
```diff
@@ -1,45 +1,67 @@
...
+            self.session = aiobotocore.get_session()
+        except Exception as e:
+            self.log.error(
+                "{} : Check credentials and configuration".format(e)
+            )
 
         # A lock per path to suppress writing to S3 when currently writing
         self.path_locks: Dict[str, Lock] = {}
         # Since locks are not pre-created,
         self.ability_to_create_path_lock = Lock()
```
willingc: Can we rename `ability_to_create_path_lock` to `path_lock_aware` or `path_lock_capable`?

bookstore/archive.py
```diff
@@ -1,47 +1,89 @@
...
+    A record to be archived is a tuple containing `filepath`, the `content` to
+    be archived, and the length of time in queue for archiving (`queued_time`).
+    """
     filepath: str
     content: str
     queued_time: float
 
 
 class BookstoreContentsArchiver(FileContentsManager):
+    """Archives notebooks to storage (S3) on save by UI or parameterized workflow
```
rgbkrk: Can just end this one with `on save`, we don't have any parametrized workflows for this (that I'm aware of, I don't know what this means here).

bookstore/archive.py
```diff
@@ -1,47 +1,89 @@
...
+
+    This class is a custom Jupyter `FileContentsManager` which holds information
+    on storage location, path to it, and file to be written there.
+    
+    Bookstore settings combine with the parent Jupyter application settings.
+    
+    A session is created for the current event loop. To write to a particular
+    path on S3, acquire a lock. After acquiring the lock, `archive`
+    authenticates using the storage service's credentials. If allowed, the
+    notebook is queued to be written to storage (i.e. S3).
```
rgbkrk: 👍

bookstore/archive.py
```diff
@@ -1,47 +1,89 @@
...
-        self.log.info("Archiving notebooks to {}".format(s3_display_path(
-            self.settings.s3_bucket, self.settings.workspace_prefix)))
+        self.log.info(
+            "Archiving notebooks to {}".format(
+                s3_display_path(self.settings.s3_bucket, self.settings.workspace_prefix)
+            )
+        )
 
-        self.session = aiobotocore.get_session()
+        # TODO: create a better check for credentials - keys and config or it will fail
```
rgbkrk: This reminds me we should better document how to use `minio` locally for development.

bookstore/archive.py
```diff
@@ -1,47 +1,89 @@
...
+        try:
+            # create a session object from the current event loop
+            self.session = aiobotocore.get_session()
+        except Exception as e:
+            self.log.error(
+                "{} : Check credentials and configuration".format(e)
+            )
```
rgbkrk: Should we bail at this point and raise the exception up further?

bookstore/archive.py
```diff
@@ -1,47 +1,89 @@
...
 
-        # A lock per path to suppress writing to S3 when currently writing
+        # a collection of locks per path to suppress writing while the path may be in use
```
rgbkrk: Much improved, thank you.

bookstore/archive.py
```diff
@@ -83,12 +136,10 @@ def run_pre_save_hook(self, model, path, **kwargs):
 
         loop = ioloop.IOLoop.current()
 
-        # Offload the archival to be scheduled to work on the event loop
+        # Offload archival and schedule write to storage with the current event loop
```
rgbkrk: A+

docs/source/configuration.rst
```diff
@@ -0,0 +1,36 @@
+Configuration
+=============
+
+
+Bookstore ettings in Jupyter configuration file
+-----------------------------------------------
+
+``~/.jupyter/jupyter_notebook_config.py`` can store commonly used settings for
```
rgbkrk: I wonder if we should be adopting setting this in `jupyter_config.py` to target the more generic config setup. What do you think @mpacer?

docs/source/usage.rst
```diff
@@ -0,0 +1,47 @@
+Usage
+=====
+
+Automatic Notebook Versioning
+-----------------------------
+
+Every *save* of a notebook creates an *immutable copy* of the notebook on object storage.
+
+To simplify implementation, we currently rely on S3 as the object store, using
+`versioned buckets <https://docs.aws.amazon.com/AmazonS3/latest/dev/Versioning.html>`_.
```
rgbkrk: 👍

docs/source/usage.rst
```diff
@@ -0,0 +1,47 @@
...
+
+Each notebook path is a namespace that an external service ties into the schedule. We archive off versions,
+keeping the path intact (until a user changes them). For example, the prefixes that could be associated with
+storage types:
+
+- Notebook in "draft" form: ``/workspace/kylek/notebooks/mine.ipynb``
+- Most recent published copy of a notebook: ``/published/kylek/notebooks/mine.ipynb``
+
+Scheduled notebooks will also be referred to by the notebook ``key``. In addition, we'll need to be able to surface
+version IDs as well.
```
rgbkrk: I just realized our language for scheduled notebooks is a little early since we haven't implemented scheduling yet. (It was aspirational README driven development from me originally).

bookstore/archive.py
mpacer: Papermill doesn't have a server or contents manager so if parameterized workflow is referring to that, then this won't be archived.

willingc: Sure thing. What I was thinking was now or down the road when a notebook is saved (whatever the oirginating source) that it will be  archived. Will end the comment at save.

docs/source/configuration.rst
mpacer: Seems reasonable. For context, the name of the config files will match for the name of the JupyterApp you are running. So if you use jupyter_notebook_config.py, it won’t be activated if you run `jupyter nteract`. 

It should be ok since it will only be effectively active on NotebookApp anyway, regardless of the JupyterApp you are running (since it’s a server extension).

I really wish there were a way to tell a JupyterApp to also inherit another JupyterApp’s config file, but that would require a new API in jupyter_core. Using the general config is probably enough for now.

willingc:
Thanks for the merge. Still have a little cleanup but will do in a second PR. Thanks @mpacer and @rgbkrk for the reviews.

PR #47, opened 2019-01-31
Cleanup settings validation

mpacer:
This does a little bit of a few things:

- makes it so settings don't validate by default (you need to know what your s3_bucket is even if you are using IAM roles)
- switch to using initialize rather than __init__ for the handler
- adds validation info onto base `api/bookstore/` endpoint
- changes logging level for testing so we see more info about the server

### DIFF
--- a/bookstore/handlers.py
+++ b/bookstore/handlers.py
@@ -7,7 +7,7 @@ from notebook.utils import url_path_join
 from tornado import web
 
 from ._version import __version__, version_info
-from .bookstore_config import BookstoreSettings
+from .bookstore_config import BookstoreSettings, validate_bookstore
 from .s3_paths import s3_path, s3_key, s3_display_path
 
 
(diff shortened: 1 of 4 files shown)


bookstore/handlers.py
```diff
@@ -32,8 +34,7 @@ def get(self):
 class BookstorePublishHandler(APIHandler):
     """Publish a notebook to the publish path"""
 
-    def __init__(self, *args, **kwargs):
-        super(APIHandler, self).__init__(*args, **kwargs)
+    def initialize(self):
```
rgbkrk: Thanks for cleaning up my badness here. 😸

bookstore/handlers.py
```diff
@@ -98,14 +99,20 @@ def __init__(self, *args, **kwargs):
...
+    base_bookstore_pattern = url_path_join(web_app.settings['base_url'], '/api/bookstore')
     web_app.add_handlers(host_pattern, [(base_bookstore_pattern, BookstoreVersionHandler)])
-
-    if not nb_app.config.get("BookstoreSettings"):
-        nb_app.log.info("Not enabling bookstore publishing since bookstore endpoint not configured")
+    bookstore_settings = BookstoreSettings(parent=nb_app)
+    web_app.settings['bookstore_validated'] = validate_bookstore(bookstore_settings)
+                                               # and nb_app.nbserver_extensions.get("bookstore")
+                                               # need to delay adding this check until server
+                                               # has been updated
```
rgbkrk: Super smart, I love this.

PR #53, opened 2019-02-01
Pin conflicting dependencies and correct autodoc

willingc:
- Pins several packages in `setup.py` and `requirements.txt` since botocore, boto3, aiobotocore, aiohttp have conflicting dependencies
- Use latest image for ReadTheDocs so we can use Python 3
- Update path for autodoc import

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -74,10 +74,6 @@ setup(
         's3fs',
         'tornado >= 5.1.1',
         'aiobotocore==0.10.0',
-        'aiohttp==3.3.2',
-        'botocore==1.12.49',
-        'boto3==1.9.49',
-        'awscli==1.16.59',
     ],
     extras_require={
         'test': [
--- a/setup.py
+++ b/setup.py
@@ -74,6 +74,10 @@ setup(
         's3fs',
         'tornado >= 5.1.1',
         'aiobotocore==0.10.0',
+        'aiohttp==3.3.2',
+        'botocore==1.12.49',
+        'boto3==1.9.49',
+        'awscli==1.16.59',
     ],
     extras_require={
         'test': [
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -18,8 +18,6 @@ import sys
 import m2r
 
 sys.path.insert(0, os.path.abspath('..'))
-#sys.path.append(os.path.join(os.path.dirname(__name__), '..'))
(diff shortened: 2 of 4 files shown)


setup.py
```diff
@@ -74,6 +74,10 @@
         's3fs',
         'tornado >= 5.1.1',
         'aiobotocore==0.10.0',
+        'aiohttp==3.3.2',
+        'botocore==1.12.49',
+        'boto3==1.9.49',
+        'awscli==1.16.59',
```
rgbkrk: The only gross thing about this is that when trying to deploy this, when we have _other_ packages with strongly pinned versions inside an environment (docker container, etc.), it's hard to resolve all of them. Our deployment strategy has been to install them in the order they allow us to be flexible with the `botocore` version. 😭

rgbkrk: @cHYzZQo (Peter), one of the leads of the image we build with recommends:

>  i would suggest not pinning
> or if you pin, be flexible about the patch release version

He can share more context about the awkwardness around botocore.

rgbkrk: At the very least, we should pin to the minor version and allow flexibility for the patch release.

willingc: Let's try this instead. Let's keep the original code not the proposed changes. I will be unable to build autodoc on RTD using pip but that's a small price to pay vs maintainability.

I'm going to try switching the docs build to using conda to build the environment on RTD.

rgbkrk: Thank you and sorry about the difficulty around this one.

willingc: No worries. There are some ugly dependencies around botocore and boto3 which are a bit surprising i.e. worst that I have seen in a while. I would be curious why there is awkwardness around botocore since it is also mentioned as challenging in the aiohttp source.

I have several workarounds in mind.

cHYzZQo: They've started doing automated builds that trigger every couple days, constantly bumping the version even when there are no code changes. So any pinning you do very quickly falls behind 'latest' leading to all sorts of conflicts between other packages that depend on them.

willingc: Thanks @cHYzZQo. Yeah, I noticed that their release cadence was way more frequent than most Python projects. aiobotocore pins these and that's part of the problem with dependency conflicts. Which makes me think that perhaps we float aiobotocore upwards `>=0.10.0` and not `pin ==` as well. Thoughts?

cHYzZQo: yeah that would be fine i think

PR #55, opened 2019-02-02
Add aioboto3 dependency and float up aiobotocore

willingc:
- Adds the [aioboto3](https://pypi.org/project/aioboto3/) package which when installed after aiobotocore resolves the boto3 dependency conflict since aioboto3 is a wrapper around boto3 and aiobotocore
- Build docs from latest RTD image so we can use Python 3.6


rgbkrk:
Thanks for digging in and resolving this.

PR #56, opened 2019-02-02
Polish bookstore configuration docstrings

willingc:


bookstore/bookstore_config.py
```diff
@@ -1,17 +1,49 @@
...
+        environment variable JPYNB_S3_ENDPOINT_URL
+    - s3_region_name : str("us-east-1")
+        environment variable JPYNB_S3_REGION_NAME
+    - s3_bucket : str("")
+        bucket name, environment variable JPYNB_S3_BUCKET
+
+    This bookstore setting determines the resources available:
+    
+    - max_threads : int(16)
+        Maximum threads from the threadpool available to do S3 read and writes
```
rgbkrk: I just realized we can deprecate max_threads since we're now on aiobotocore.

PR #65, opened 2019-02-25
change API for validation to return an dict with feature specific checks

mpacer:
Per conversation with @stormpython about feature specific config and settling on an API as soon as possible, this switches over the current validation code over to returning a dict (object once converted to json).

- "bookstore_validation" Checks whether bookstore overall has been 
validated (is false by default unless you set a s3_bucket)
- "archive_validation" Checks whether archiving has been configured
- "published_validation" Checks whether publishing has been configured


rgbkrk:
I think this is fit to ship and start using. Want to make a minor release?

mpacer:
Sure I'll do the release.

Issue #64, opened 2019-02-15
Landing page for cloning that has a confirmation dialog before saving


mpacer:
To mitigate Cross Site Request Forgery attacks, create a landing page for cloning (e.g. at `/bookstore/clone`) that will provide a simple confirmation dialog before cloning with the API in #63 followed by redirecting the user to the notebook in nteract.

PR #66, opened 2019-02-26
WIP: Clone endpoint

mpacer:
This is a first attempt at setting up a clone endpoint 

- creates new publish module to house the publish handler
- updates docs for testing
- creates new clone module to house the clone handler
   - iterates through API designs while providing example queries for testing at each stage

Final design is

```
<nb_origin>/api/bookstore/clone/("/?s3_bucket=nteract-notebooks&s3_key=published/whateverwewant.json"
```

Note as of right now there's no good way to specify configuration for multiple buckets. If we want to enable that, we may need to iterate on how to achieve that. 

Right now this works even without any keys being passed into the session explicitly because I have a local `~/.aws/credentials` file. We could choose to only support that. The example queries I provided require that you have access to the "nteract-notebooks" s3 bucket.


bookstore/publish.py
```diff
@@ -0,0 +1,75 @@
...
+
+        if path == '' or path == '/':
+            raise web.HTTPError(400, "Must have a path to publish to")
+
+        model = self.get_json_body()
+
+        if model:
+            await self._publish(model, path.lstrip('/'))
+        else:
+            raise web.HTTPError(400, "Cannot publish empty model")
```
rgbkrk: Thanks for extracting publish out as a separate handler!

ci/clone_request.py
```diff
@@ -0,0 +1,13 @@
...
+
+def main(queries):
+    return requests.get(f"http://localhost:8888/api/bookstore/clone{queries}")
+
+
+if __name__ == "__main__":
+    res = main("?s3_bucket=nteract-notebooks&s3_key=published/whateverwewant.json")
+    print(res.json())
+    res = main("/?s3_bucket=nteract-notebooks&s3_key=published/whateverwewant.json")
+    print(res.json())
```
rgbkrk: I assume this only works with tokens off here?

package.json
```diff
@@ -8,7 +8,8 @@
   },
   "private": "true",
   "scripts": {
-    "test": "node ci/integration.js"
+    "test": "node ci/integration.js",
+    "test:server": "./ci/local.sh"
```
rgbkrk: Did you mean to add in `ci/clone_request.py` somewhere in here too?

README.md
```diff
@@ -81,3 +81,11 @@ c.BookstoreSettings.s3_bucket = "<bucket-name>"
...
 c.BookstoreSettings.s3_secret_access_key = <AWS Secret Access Key / IAM Secret Access Key>
 ```
+
+## Developing
+
+If you are developing on bookstore you will want to run the ci tests locally and to make releases.
+
+Use [CONTRIBUTING.md](./CONTRIBUTING.md) to learn more about contributing.
+Use [running_ci_locally.md](./running_ci_locally.md) to learn more about running ci tests locally. 
+Use [RELEASING.md](./RELEASING.md) to learn more about releasing bookstore.
```
rgbkrk: 👍

bookstore/clone.py
```diff
@@ -0,0 +1,66 @@
...
+        resp_content = {"s3_path": full_s3_path, "content": content.decode('utf-8')}
+
+        self.log.info(obj)
+        if 'VersionId' in obj:
+            resp_content["versionID"] = obj['VersionId']
+
+        resp_str = json.dumps(resp_content)
+        self.finish(resp_str)
+
+    async def get(self):
```
rgbkrk: This is where we'll need to render a template that forces the user to make a choice between cloning and not cloning. https://github.com/nteract/bookstore/issues/64

```
self.render_template('clone.html')
```

bookstore/clone.py
```diff
@@ -0,0 +1,66 @@
...
+        That's ok! 
+        But this all should go away.
+        """
+        s3_bucket = self.get_argument("s3_bucket", "")
+        file_key = self.get_argument("s3_key", "")
+        self.log.info("About to clone from %s", file_key)
+
+        if file_key == '' or file_key == '/':
+            raise web.HTTPError(400, "Must have a key to clone from")
+        await self._clone(s3_bucket, file_key)
```
rgbkrk: We should either remove this before merging or for now just render a basic page. If you go ahead and create `clone.html` (in this same directory), you can then `self.write(self.render_template('clone.html'), {})`. `clone.html` can just be a page that only says "Cloning not implemented yet". Then we can follow it up with the rest in a new PR.

package.json
mpacer: I had only been doing this interactively, I was going to get the functionality laid out before I moved forward with that.

ci/clone_request.py
mpacer: yes i mentioned this in the commit messages expanded version. I don't intend for this to be the longterm solution — it is just a way to rapidly prototype these things.

bookstore/publish.py
mpacer: Yay refactors!

package.json
mpacer: I added more documentation in baef651 to explain the purpose of this file. @stormpython Do you think it explains it thoroughly enough?

bookstore/clone.py
mpacer: Addressed with 6875936.

bookstore/clone.py
mpacer: Addressed with 6875936.

mpacer:
In conversation with @yakticus, @cHYzZQo, & @stormpython, I'm going to re-work some of this.

This will involve 

- the POST endpoint should do the cloning rather than the GET endpoint 
- the GET endpoint will need to still exist to avoid xsrf protection issues
- the GET endpoint will have a form that triggers the POST endpoint
- the API should act like a resource not a verb and so should be `api/bookstore/cloned` not `api/bookstore/cloned`

package.json
stormpython: Yes, the comments are sufficient.

ci/jupyter.js
```diff
@@ -6,12 +6,41 @@ const { sleep } = require("./sleep");
 global.XMLHttpRequest = require("xmlhttprequest").XMLHttpRequest;
 const { ajax } = require("rxjs/ajax");
 
+let url_path_join = function(...pieces) {
```
stormpython: It would be good to move this function to a utility file as per our discussion.

stormpython: Actually, may we should consider importing [url-join](https://www.npmjs.com/package/url-join).

mpacer:
Closing this PR because it is causing more confusion than help at this point and is made obsolete by #75

PR #67, opened 2019-02-27
switches from storing the validation to storing bookstore settings

mpacer:
This is a further iteration on how we will surface the settings via api and via the webapp settings for use by other serverextensions.

### DIFF
--- a/bookstore/handlers.py
+++ b/bookstore/handlers.py
@@ -21,7 +21,15 @@ class BookstoreVersionHandler(APIHandler):
 
     @web.authenticated
     def get(self):
-        self.finish(json.dumps({"bookstore": True, **self.settings['bookstore']}))
+        self.finish(
+            json.dumps(
+                {
+                    "bookstore": True,
+                    "version": self.settings['bookstore']["version"],
+                    "validation": self.settings['bookstore']["validation"],
+                }
+            )
+        )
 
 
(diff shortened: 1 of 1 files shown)


bookstore/handlers.py
```diff
@@ -21,15 +21,7 @@ class BookstoreVersionHandler(APIHandler):
...
-        self.finish(
-            json.dumps(
-                {
-                    "bookstore": True,
-                    "version": version,
-                    "bookstore_validation": self.settings['bookstore_validation'],
-                }
-            )
-        )
+        self.finish(json.dumps({"bookstore": True, **self.settings['bookstore']}))
```
rgbkrk: ~~We should still expose the version on the API endpoint for our own debugging. Or wait... does `self.settings['bookstore']` have this?~~ I see, that's below.

Is there anything in `self.settings['bookstore']` that shouldn't be leaked out as part of the payload? One code smell I see here is that if anyone adds a new property to the settings, it's not going to be obvious that it will get leaked out onto the API endpoint here. I think this should still use specific keys.

mpacer: @stormpython I think e54de64 addresses @rgbkrk's concerns and so as long as it looks good to you I think this PR is good to go.

mpacer:
@stormpython is there anything more you think needs to be done for this to be merged?

PR #69, opened 2019-03-15
Clients!

mpacer:
# Publish your notebook through IPython

This introduces two kinds of clients `NotebookClient` and `BookstoreClient`.  These are intended for use inside Jupyter notebooks. 

Furthermore, they use a couple of tricks with IPython and the `notebook` library to identify which kernel, session, and notebook you are in.

This enables you to do things like 

```
from bookstore.client.bookstore import BookstoreClient
bookstore = BookstoreClient()
bookstore.publish()
```

As a way to publish the current notebook. 

Lots of functionality could be added onto what I have here… but I wanted to start with the minimal set of methods to accomplish the purpose of identifying which kernel/notebook/server you were in. 

## Caveats

- You need to separately save before you can publish. There is no general way I know of to trigger a save from the backend.

- This will only work appropriately if you have only one notebook attached to each kernel. This is the standard use case, so I think that's an ok problem. 

- This will only work in IPython.

- This _will not_ work in nbconvert, papermill, or even jupyter_client. But that also kinda makes sense… since in none of those cases do you have a notebook or bookstore server for which the client would exist.

I'll need to improve docs on this but omg this is so nice to use!
 @willingc @stormpython @rgbkrk

### DIFF
--- a/bookstore/client/notebook.py
+++ b/bookstore/client/notebook.py
@@ -99,7 +99,8 @@ class NotebookClient:
 
     def get_sessions(self):
         target_url = f"{self.sessions_endpoint}"
-        resp = requests.get(target_url)
+        headers = {'Authorization': f'token {self.token}'}
+        resp = requests.get(target_url, headers=headers)
         return resp.json()
 
     @property
@@ -109,7 +110,8 @@ class NotebookClient:
 
     def get_kernels(self):
         target_url = f"{self.sessions_endpoint}"
-        resp = requests.get(target_url)
+        headers = {'Authorization': f'token {self.token}'}
+        resp = requests.get(target_url, headers=headers)
         return resp.json()
 
     @property
@@ -119,7 +121,8 @@ class NotebookClient:
 
     def get_contents(self, path):
         target_url = f"{self.contents_endpoint}{path}"
-        resp = requests.get(target_url)
+        headers = {'Authorization': f'token {self.token}'}
+        resp = requests.get(target_url, headers=headers)
         return resp.json()
 
 
@@ -160,4 +163,3 @@ class CurrentNotebookClient(NotebookClient):
     @property
     def kernel_id(self):
         return os.path.basename(self.connection_file).lstrip('kernel-').rstrip('.json')
-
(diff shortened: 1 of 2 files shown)


bookstore/client/notebook.py
```diff
@@ -0,0 +1,165 @@
+# # Building a Notebook client
```
willingc: Let's move this block from Line 1 - 19 into a file docstring.

bookstore/client/notebook.py
```diff
@@ -0,0 +1,165 @@
...
+    password: bool
+    pid: int
+    port: int
+    secure: bool
+    token: str
+    url: str
+
+
+class KernelInfo:
+    #     id: str # 'f92b7c8b-0858-4d10-903c-b0631540fb36',
```
willingc: Let's make this a class docstring.

mpacer: Agreed! That sounds good.

mpacer:
@willingc A lot of the classes are to act as types (if we could assume python 3.7 I'd have made them data classes). The reason they aren't just inheriting from `NamedTuple` is that I couldn't figure out how to get `NamedTuple` to do the appropriate transformation on the lower level class (e.g., `KernelInfo`) based on the typing that I was providing at the class attribute level.

bookstore/client/notebook.py
```diff
@@ -0,0 +1,165 @@
...
+
+class NotebookClientCollection:
+    nb_client_gen = lambda: (NotebookClient(x) for x in list_running_servers())
+    sessions = {x.url: x.sessions for x in nb_client_gen()}
+
+    @classmethod
+    def current_server(cls):
+        for server_url, session_dict in cls.sessions.items():
+            for session_id, session in session_dict.items():
+                python_compat_session(session)
```
mpacer: oops — I had actually intended to make a copy of the session there, though now in retrospect I don't need to so I can clean that up.

mpacer:
I also want to surface cloning from this API. I also was thinking this might make it easier to create unit tests for the intermediate data types (since they were otherwise implicit).

willingc:
@mpacer I haven't forgotten this. I've been swamped. Will try to get it loaded this weekend and think more about it. Feel free to merge if you are blocked on this and we can always refactor.

PR #70, opened 2019-03-23
Refactor docstrings and docs for bookstore modules

willingc:


bookstore/bookstore_config.py
```diff
@@ -68,18 +76,15 @@ class BookstoreSettings(LoggingConfigurable):
 
 
 def validate_bookstore(settings: BookstoreSettings):
-    """Ensure that bookstore settings are valid to appropriately set up the application.
+    """Validate bookstore configuration settings.
```
mpacer: I like the wording change; I think it would help to clarify that what's being validated is not the types but whether they are configured in a way to successfully use Bookstore.

PR #71, opened 2019-03-23
Refactor naming and expand client docstrings

willingc:
- Adds/edits docstrings for clients
- update filenames to avoid confusion
~~- reorder classes in notebook client file for maintainability~~
- add autodoc doc pages


willingc:
@mpacer let's keep iterating on the clients

bookstore/client/nb_client.py
```diff
@@ -126,30 +73,101 @@ def get_contents(self, path):
...
+       -------
+       kernel_id: str  # 'f92b7c8b-0858-4d10-903c-b0631540fb36',
+       name: str  # 'dev',
+       last_activity: str  #'2019-03-14T23:38:08.137987Z',
+       execution_state: str  #'idle',
+       connections: int  # 0 connections to the kernel
+    """
+
+    def __init__(self, kernel_id, name, last_activity, execution_state, connections):
+        # TODO: verify that *args is not needed
```
mpacer: Perhaps it should have been just a * not a *args, I was intending to make these keyword only arguments with no default values.

bookstore/client/nb_client.py
```diff
@@ -126,30 +73,101 @@ def get_contents(self, path):
...
 
     @classmethod
     def current_server(cls):
+        """class method for current notebook server"""
         for server_url, session_dict in cls.sessions.items():
             for session_id, session in session_dict.items():
-                python_compat_session(session)
-                if NotebookSession(**session).kernel.id == extract_kernel_id(
+                python_compatible_session(session)
+                if NotebookSession(**session).kernel.kernel_id == extract_kernel_id(
```
mpacer: I feel mixed about the idea that we're having the accessor being different than the spec's response. I totally agree that it's clearer though from the perspective of distinguishing from the session's id.

bookstore/client/nb_client.py
```diff
@@ -126,30 +73,101 @@ def get_contents(self, path):
...
     sessions = {x.url: x.sessions for x in nb_client_gen()}
 
     @classmethod
     def current_server(cls):
+        """class method for current notebook server"""
         for server_url, session_dict in cls.sessions.items():
             for session_id, session in session_dict.items():
-                python_compat_session(session)
-                if NotebookSession(**session).kernel.id == extract_kernel_id(
+                python_compatible_session(session)
```
mpacer: we probably can remove this?

bookstore/client/nb_client.py
```diff
@@ -126,30 +73,101 @@ def get_contents(self, path):
...
+       kernel_id: str  # 'f92b7c8b-0858-4d10-903c-b0631540fb36',
+       name: str  # 'dev',
+       last_activity: str  #'2019-03-14T23:38:08.137987Z',
+       execution_state: str  #'idle',
+       connections: int  # 0 connections to the kernel
+    """
+
+    def __init__(self, kernel_id, name, last_activity, execution_state, connections):
+        # TODO: verify that *args is not needed
+        self.kernel_id = kernel_id
```
mpacer: Even if we go back to id for the parameter, I'm not sure whether we want it to be `id` or `kernel_id` on the object.

I was imagining this as being basically a typed version of the object we're getting back. So, from that perspective, we shouldn't be changing the name in either case (also discussed elsewhere).

bookstore/client/nb_client.py
```diff
@@ -1,94 +1,41 @@
...
     def __init__(self, nb_config):
         self.nb_config = nb_config
-        self.nb_record = LiveNotebookRecord(**self.nb_config)
-        self.url = self.nb_record.url.rstrip(
-            "/"
-        )  # So that we can have full API endpoints without double //
-        self.token = self.nb_record.token
+        self.nbserver_record = LiveNotebookServerRecord(**self.nb_config)
+        # Right strip '/' so we can have full API endpoints without double '//'
+        self.url = self.nbserver_record.url.rstrip("/")
```
mpacer: we might be able to reuse a notebook utility for this functionality. I'll look into it.

bookstore/client/nb_client.py
```diff
@@ -1,94 +1,41 @@
...
-#   'port': 8888,
-#   'secure': False,
-#   'token': '',
-#   'url': 'http://localhost:8888/'}]
-# ```
-
+"""Client to test bookstore endpoints from within a notebook.
+
+
+TODO: (Clarify) We want to test our bookstore endpoints, but it's no fun having
```
mpacer: It ends up being not just for testing, but this is a good todo to have :)

willingc:
@mpacer @MSeal I've gone back and simplified this PR to docstring changes, a missed import, autodoc of files, and black autoformat. Let's go ahead and merge. Then we can begin tackling the other PR. We should decide whether to do a phone call review or break it up into smaller PRs. Thanks.

PR #75, opened 2019-04-06
Clone client

mpacer:
This is the most basic version of the cloning endpoint for bookstore.

This adds:
- cloning functionality to the bookstore client (the `BookstoreClient` enables accessing bookstore functionality from inside a notebook)
- POST endpoint (actually does the cloning… uses Jupyter’s POST contents api to handle saving)
- GET endpoint (allows getting around xsrf issues when cloning needs to be triggered from outside of a notebook context; e.g., from commuter)
- further extensions to the minio based mock-s3 integration testing suite
- more documentation around how to setup and use that testing suite

This incorporates a lot of the work from #66 and renders it obsolete.


bookstore/clone.html
```diff
@@ -0,0 +1,56 @@
...
+            }
+          };
+          xhr.onerror = function(e) {
+            console.error(xhr.statusText);
+          };
+
+          xhr.open("POST", url, true);
+          xhr.withCredentials = true;
+          xhr.setRequestHeader("X-XSRFToken", xsrf_token);
+          xhr.send(post_json);
```
mpacer: @stormpython I'm sure I'm approaching this in a unnecessarily complicated way and would really appreciate any comments on how to simplify this (note… it doesn't work right now at all… :D)

mpacer: Ah — I got it working in the latest iteration.

mpacer:
So the biggest open question now for me is how we should approach testing cloning. I don't want to start up selenium to test this… but there's not another good way to ensure that the cookie is being stored appropriately in the browser as far as I can think of it at least. Since we're already running in node… maybe we could leverage that? I'm not sure.

ci/integration.js
```diff
@@ -5,7 +5,7 @@ const fs = require("fs");
 const _ = require("lodash");
```
stormpython: I would break out the `originalNotebook` into its own file. It makes it easier to make changes to the notebook and typically these mocked objects are usually in their own file if not folder.

ci/jupyter.js
```diff
@@ -6,12 +6,41 @@ const { sleep } = require("./sleep");
...
 
+let url_path_join = function(...pieces) {
+  // """Join components of url into a relative url
+  // Use to prevent double slash when joining subpath. This will leave the
+  // initial and final / in place
+  // """
+  initial = pieces[0].startsWith("/");
+  final = pieces[pieces.length - 1].endsWith("/");
+  const hi = [];
+  for (s of pieces) {
```
stormpython: Did you know you can use `Array.map()` to achieve the same thing here? Although, I do believe a `for` loop is more performant. Here is a [blog post](https://medium.com/@ExplosionPills/map-vs-for-loop-2b4ce659fb03) that talks about using `for` vs `map`. Again, not required. But generally `map` is used in replace of a `for` loop.

ci/jupyter.js
```diff
@@ -6,12 +6,41 @@ const { sleep } = require("./sleep");
 global.XMLHttpRequest = require("xmlhttprequest").XMLHttpRequest;
 const { ajax } = require("rxjs/ajax");
 
+let url_path_join = function(...pieces) {
```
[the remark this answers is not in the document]
stormpython: In general, `const` should be used instead of `let` if the variables value never changes.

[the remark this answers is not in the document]
mpacer: Fair enough!

ci/integration.js
mpacer: Should this be in it's own PR focusing on the ci integration tests? I can do it in this one, but it's already a PR with a pretty wide scope.

mpacer: NB: the other changes you asked for are stuff I'm introducing so I have no issue including those fixes in this PR.

stormpython: Yes, it should go into its own PR. You can just marked this comment as resolved.

bookstore/clone.html
```diff
@@ -0,0 +1,63 @@
...
+    <script>
+      window.onload = function() {
+        var button = document.getElementById("mybutton");
+        button.onclick = function(event) {
+          var post_json = {
+            s3_bucket: "{{ s3_bucket }}",
+            s3_key: "{{ s3_key }}"
+          };
+          var clone_url = "{{ base_uri }}/api/bookstore/cloned/";
+          var r = document.cookie.match("\\b" + "_xsrf" + "=([^;]*)\\b");
```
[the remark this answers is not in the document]
mpacer: This is the literal code that we use from jupyter notebook:

```javascript
    var _get_cookie = function (name) {
        // from tornado docs: http://www.tornadoweb.org/en/stable/guide/security.html
        var r = document.cookie.match("\\b" + name + "=([^;]*)\\b");
        return r ? r[1] : undefined;
    }
```

https://github.com/jupyter/notebook/blob/b8b66332e2023e83d2ee04f83d8814f567e01a4e/notebook/static/base/js/utils.js#L800-L804

So this logic would need to live within `getCookieByName` and would need to live in this script tag since we're not packaging this at all. Would that make you feel more confident?

ci/jupyter.js
```diff
@@ -6,12 +6,41 @@ const { sleep } = require("./sleep");
...
+  }
+  return result;
+};
 class JupyterServer {
   constructor(config = {}) {
     this.port = config.port || 9988;
     this.ip = config.ip || "127.0.0.1";
     this.scheme = config.scheme || "http";
     this.token = null;
+    // this.baseUrl = config.baseUrl || "";
```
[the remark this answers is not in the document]
mpacer: I don't remember why I added it twice, I'll get rid of this one.

ci/jupyter.js
mpacer: Sounds good!

ci/jupyter.js
```diff
@@ -77,8 +107,9 @@ class JupyterServer {
   async writeNotebook(path, notebook) {
     // Once https://github.com/nteract/nteract/pull/3651 is merged, we can use
     // rx-jupyter for writing a notebook to the contents API
+    const apiPath = "/api/contents/";
     const xhr = await ajax({
-      url: `${this.endpoint}/api/contents/${path}`,
+      url: url_path_join(this.endpoint, apiPath, path),
```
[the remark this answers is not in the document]
mpacer: While developing this is constantly an issue.

Yes, its developer error with consistency around whether routes are defined with `/` at the beginning & end or not. Trying to keep track of it leaves it easy to mess up. 

I'm bringing this pattern over from the jupyter notebook server, where it is used extensively. We're already importing it for the serverside code.

This a huge quality of life improvement, trust me.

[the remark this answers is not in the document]
mpacer: For reference to the server side code: 

```python
def url_path_join(*pieces):
    """Join components of url into a relative url
    Use to prevent double slash when joining subpath. This will leave the
    initial and final / in place
    """
    initial = pieces[0].startswith('/')
    final = pieces[-1].endswith('/')
    stripped = [s.strip('/') for s in pieces]
    result = '/'.join(s for s in stripped if s)
    if initial: result = '/' + result
    if final: result = result + '/'
    if result == '//': result = '/'
    return result
```

https://github.com/jupyter/notebook/blob/master/notebook/utils.py#L48-L61

mpacer:
I think I've addressed your comments on the js side @captainsafia @stormpython. 

@willingc @MSeal would either of you be able to take a look at the python side of things?

ci/utils.js
```diff
@@ -0,0 +1,30 @@
+const url_path_join = function(...pieces) {
+  // """Join components of url into a relative url
+  // Use to prevent double slash when joining subpaths.
+  // This will leave the initial and final / in place
+  //
+  // url_path_join("http://127.0.0.1:9988", "mybaseUrl/ipynb", "/api/contents//") => "http://127.0.0.1:9988/mybaseUrl/ipynb/api/contents/"
+  //
+  // They will be readded in between words, and at the beginning and end if they were
+  // """
+  initial = pieces[0].startsWith("/");
```
stormpython: missing `const`

ci/utils.js
```diff
@@ -0,0 +1,30 @@
...
+  final = pieces[pieces.length - 1].endsWith("/");
```
stormpython: missing `const`

ci/utils.js
```diff
@@ -0,0 +1,30 @@
...
+  const hi = pieces
+    .filter(el => el !== "")
+    .map(el => el.replace(/(^[ /]+)|([/]+$)/g, "")); //[];
+  result = hi.join("/");
```
stormpython: This can be rewritten to:

```javascript
const result = pieces
    .filter(el => el !== "")
    .map(el => el.replace(/(^[ /]+)|([/]+$)/g, ""))
    .join("/");
```

bookstore/client/notebook.py
```diff
@@ -58,28 +62,48 @@ class KernelInfo:
...
     #     type: str #'notebook',
     #     kernel: KernelInfo
-    #     notebook: dict # {'path': 'Untitled38.ipynb', 'name': ''}}}
+    #     notebook: dict # deprecated API {'path': 'Untitled38.ipynb', 'name': ''}}}
 
     def __init__(self, *args, path, name, type, kernel, notebook, **kwargs):
+        self.model = {
+            "path": path,
+            "name": name,
+            "type": type,
```
MSeal: we should avoid naming variables with reserved names like `type`

bookstore/clone.py
```diff
@@ -0,0 +1,112 @@
...
+    async def _clone(self, s3_bucket, file_key):
+        path = file_key
+
+        self.log.info(f"bucket: {s3_bucket}")
+        self.log.info(f"key: {file_key}")
+        full_s3_path = s3_path(s3_bucket, path)
+
+        async with self.session.create_client(
+            's3',
+            aws_secret_access_key=self.bookstore_settings.s3_secret_access_key,
```
MSeal: Do we force the config to be loaded into `bookstore_settings`? There's many instances where these credentials are fetch JIT from the system config loader. Likely a conversation out of scope of this PR though.

bookstore/clone.py
```diff
@@ -0,0 +1,112 @@
...
+        The payload type for the request should be 
+        {
+          "s3_bucket": string,
+          "s3_key": string,
+          "target_path"?: string
+        }
+        The response payload should match the standard Jupyter contents API POST response.
+        """
+        model = self.get_json_body()
+        s3_bucket = model.get("s3_bucket", "")
```
MSeal: This seems wrong, an empty string bucket if the key isn't found would blow up later in the clone call. Shouldn't it fail fast here?

bookstore/__init__.py
```diff
@@ -1,3 +1,9 @@
+import os
+
+PACKAGE_DIR: str = os.path.realpath(os.path.dirname(__file__))
+
+del os
```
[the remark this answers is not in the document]
MSeal: Maybe not expose internal paths to the init of the package as well ?

bookstore/publish.py
```diff
@@ -0,0 +1,80 @@
...
+class BookstorePublishHandler(APIHandler):
+    """Publish a notebook to the publish path"""
+
+    def initialize(self):
+        """Initialize a helper to get bookstore settings and session information quickly"""
+        self.bookstore_settings = BookstoreSettings(config=self.config)
+        self.session = aiobotocore.get_session()
+
+    @web.authenticated
+    async def put(self, path=''):
```
MSeal: Why have the default if it will 400 anyway? Is this needed for the @web decorator?

bookstore/publish.py
```diff
@@ -0,0 +1,80 @@
...
+            endpoint_url=self.bookstore_settings.s3_endpoint_url,
+            region_name=self.bookstore_settings.s3_region_name,
+        ) as client:
+            self.log.info("Processing published write of %s", path)
+            obj = await client.put_object(
+                Bucket=self.bookstore_settings.s3_bucket, Key=file_key, Body=json.dumps(content)
+            )
+            self.log.info("Done with published write of %s", path)
+
+        self.set_status(201)
```
MSeal: should this be the last line before `finish`?

bookstore/client/bookstore.py
```diff
@@ -4,18 +4,35 @@
...
-    def publish(self):
-        nb_json = self.get_contents(self.notebook['path'])['content']
-        req = {
-            "json": {"type": "notebook", "content": nb_json},
-            "headers": {"Content-Type": "application/json", 'Authorization': f'token {self.token}'},
-        }
+    def publish(self, path=None):
+        if path is None:
+            path = self.session.path
+        nb_json = self.get_contents(self.session.path)['content']
```
willingc:
```suggestion
        nb_json = self.get_contents(path)['content']
```

bookstore/client/bookstore.py
```diff
@@ -4,18 +4,35 @@
...
+        json_body = {"type": "notebook", "content": nb_json}
+
+        target_url = f"{self.publish_endpoint}{self.session.path}"
```
willingc:
```suggestion
        target_url = f"{self.publish_endpoint}{path}"
```

bookstore/client/bookstore.py
```diff
@@ -4,18 +4,35 @@
...
+    @property
+    def clone_endpoint(self):
+        api_endpoint = "/api/bookstore/cloned/"
+        return f"{self.url}{api_endpoint}"
 
-        resp = requests.put(target_url, **req)
+    def clone(self, s3_bucket="", s3_key="", target_path=""):
+        s3_bucket = s3_bucket or self.default_bucket
+        json_body = {"s3_bucket": s3_bucket, "s3_key": s3_key, "target_path": target_path}
+        target_url = f"{self.clone_endpoint}"
```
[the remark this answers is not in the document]
willingc: Better style to not create the target_url variable which is only used in the next line. Instead put `f"{self.clone_endpoint}" in the post request.

bookstore/client/bookstore.py
```diff
@@ -4,18 +4,35 @@
...
+
+        target_url = f"{self.publish_endpoint}{self.session.path}"
+
+        resp = self.req_session.put(target_url, json=json_body)
+        return resp
 
-        target_url = f"{self.publish_endpoint}{self.notebook['path']}"
+    @property
+    def clone_endpoint(self):
+        api_endpoint = "/api/bookstore/cloned/"
```
willingc: cloned feels a bit off on the api to me. I would prefer a noun: clone, copy, content...

bookstore/client/notebook.py
```diff
@@ -182,12 +204,11 @@ def __init__(self):
...
-        self.notebook = NotebookSession(**self.session).notebook
 
     @property
     def connection_file(self):
         return get_ipython().parent.parent.connection_file
 
     @property
     def kernel_id(self):
-        return os.path.basename(self.connection_file).lstrip('kernel-').rstrip('.json')
+        return extract_kernel_id(self.connection_file)
```
willingc: Do we anticipate that this function (extract_kernel_id) will be used more than once? If not, it may be better to leave as is.

bookstore/clone.py
```diff
@@ -0,0 +1,112 @@
...
+            resp_content["versionID"] = obj['VersionId']
+        return model
+
+    @web.authenticated
+    async def get(self):
+        """Open a page that will allow you to clone a notebook from a specific bucket.
+        """
+        s3_bucket = self.get_argument("s3_bucket", "")
+        # s3_paths module has an s3_key function; file_key avoids confusion
+        file_key = self.get_argument("s3_key", "")
```
willingc: Perhaps `s3_file_key`

bookstore/publish.py
willingc:
```suggestion
    async def put(self, path):
```

bookstore/publish.py
```diff
@@ -0,0 +1,80 @@
...
+        if model:
+            await self._publish(model, path.lstrip('/'))
+        else:
+            raise web.HTTPError(400, "Cannot publish an empty model")
+
+    async def _publish(self, model, path):
+        """Publish notebook model to the path"""
+        if model['type'] != 'notebook':
+            raise web.HTTPError(400, "bookstore only publishes notebooks")
+        content = model['content']
```
willingc: I would skip this assignment and use `model['content']` when doing the put later.

bookstore/__init__.py
```diff
@@ -1,3 +1,9 @@
+import os
```
willingc:
```suggestion
```

bookstore/__init__.py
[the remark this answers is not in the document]
willingc: Yes let's remove this.

willingc:
I'm going to go ahead and merge this as is. I'll open a followup PR to address any outstanding feedback. (I'm bored with pneumonia and need something to do.)

bookstore/clone.html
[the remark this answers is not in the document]
willingc: See PR #88

ci/jupyter.js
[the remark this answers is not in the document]
willingc: See PR #88

ci/utils.js
willingc: See PR #88

ci/utils.js
willingc: See PR #88

ci/utils.js
willingc: See PR #88

CHANGELOG.md
```diff
@@ -1,10 +1,35 @@
 # Change Log
 
+## 2.3.0
+
+[2.3.0 on Github](https://github.com/nteract/bookstore/releases/tag/2.3.0)
+
+### Significant changes
+
+Validation information is now exposed as a dict at the /api/bookstore endpoint.
```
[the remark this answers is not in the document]
willingc: See PR #87

CHANGELOG.md
```diff
@@ -1,10 +1,35 @@
...
+    }
+```
+ 
+## Versions previous to 2.3.0
+
+[2.2.1](https://github.com/nteract/bookstore/releases/tag/2.2.1)
+
+[2.2.0](https://github.com/nteract/bookstore/releases/tag/2.2.0)
 
 [Unreleased](https://github.com/nteract/bookstore/compare/0.2.1...HEAD)
```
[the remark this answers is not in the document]
willingc: See PR #87

MSeal:
@willingc I think we maybe shouldn't have merged as is. I'd of preferred some tests and author responses to a few of the comments in the PR.

willingc:
Sorry, @MSeal. If you or @mpacer feel strongly about this, please feel free to revert.

PR #79, opened 2019-05-09
Correct the doc build failure

### LINKED ISSUE #78: Correct failing doc build on master
https://readthedocs.org/projects/bookstore/builds/

It looks like the install step of a dev version is failing.

willingc:
- Update RTD config file to use version 2 api
- Use Sphinx v2 or higher
- Add boto3 and botocore for docs build

Closes #78


requirements.txt
```diff
@@ -5,4 +5,6 @@ notebook
 s3fs
 tornado >= 5.1.1
 aioboto3
-aiobotocore>=0.10.0
+aiobotocore>=0.10.2
+boto3==1.9.91
```
MSeal: Is there a reason we're tightly pinning here? This will cause build issues for many systems that control boto version across many projects.

setup.py
```diff
@@ -73,10 +73,13 @@
         'notebook',
         's3fs',
         'tornado >= 5.1.1',
-        'aiobotocore>=0.10.0',
+        'aiobotocore==0.10.2',
```
MSeal: Any reason we don't read this from the requirements.txt? This requirement is different here and I have found that having duplicate dependency source files causes this problem often.

requirements.txt
willingc: There are some dependency issues among aioboto3, aiobotocore, boto3, and botocore. I can't seem to get the dependencies to resolve (needed for autodoc) without pinning. That said, I agree with you this is less than ideal, and I'm all ears if you have a workaround.

setup.py
willingc: Oversight that they diverged. Do we want requirements.txt or only in setup.py?

MSeal: Personally I like the `requirement.txt` as a source of truth. It's clean and simple for someone to find the requirements. I've done this in papermill and scrapbook as well. Doesn't have to be done in this PR, and so long as we collapse to one place it should be fine.

requirements.txt
MSeal: Ahh ok. So I talked with our image build team and they're super frustrated by the current boto release process. To the point that they ignore pins on boto related projects to even get anything to build in tandem.

There's a few ways we can approach this.
1) Only pin for the autodoc build tool. Make a more specific requirements-doc.txt which pins more specifically.
2) Make the boto imports lazy (inside classes) so we don't need boto for doc builds at all.

I think 1) is simpler for now and keeps the tool from aggressive pinning

setup.py
willingc: Sounds good. Looks like bumping to Sphinx 2+ and unpinning resolved. I have not idea why but 🤷‍♀

willingc: @MSeal spoke too soon. Needed to wipe and restart the rtd container. Darn dependencies.

requirements.txt
willingc: I will try 1 in the next few days. Agree with the image build team.

willingc:
[Rendered docs](https://test-bookstore.readthedocs.io/en/doc-latest/reference/archive.html) passing now and properly building autodoc pages.

PR #81, opened 2019-05-10
Polish documentation

willingc:
- Edit doc files
- add changelog release dates


mpacer:
Removing space from the end of lines in markdown files can cause the lines to collapse onto each other (being interpreted as a soft line break when in actuality it was intended to be a hard-line break). https://spec.commonmark.org/0.28/#hard-line-breaks explains this in greater depth.

willingc:
@mpacer Good point. Are there any use cases that you see where we would use multispaces at the end of a line vs a "return and blank line"? I would prefer explicit over implicit.

PR #82, opened 2019-05-10
Add pre-commit hooks

willingc:
- black for autoformatting
- flake8
- isort for imports
- pre-commit miscellaneous


MSeal:
Need to put these in more nteract projects!

mpacer:
I really disprefer precommit hooks. When using an external client (e.g., Atom) it makes it really difficult to know what you are committing. Having it format on save is much cleaner (if someone less reliable).

mpacer:
I would have preferred we keep the notes about the different categories of config https://github.com/nteract/bookstore/pull/82/files#diff-f655c6dbc41be5eac0a2fbad194e7110L18 as it explains what the different kinds of configuration do. Is there a reason to get rid of them?

If it's due to the RST headers level characters expected by sphinx, I think we can probably fix that to make it more consistent.

But I'm open to dropping them if there is a reason that the additional info is messing with formatting or something.

willingc:
> I really disprefer precommit hooks. When using an external client (e.g., Atom) it makes it really difficult to know what you are committing. Having it format on save is much cleaner (if someone less reliable).

If we add tox we can specify that the precommit hooks run with show_diff_on_failure enabled. This way one would have to do a conscious changes, add, and commit.

MSeal:
Another pattern is to only run the pre-commit hook described here during releases.

I'm definitely in the camp of pro systems cleaning up trivial changes that users may PR in without bugging those users all the time. Relying on everyone to have the same editor settings is, as mentioned, pretty unreliable.

Likely I'll start this conversation in a repo where there's more contributors, like papermill. For now given that there's only a select few primary contributors being more conservative is fine imo.

PR #88, opened 2019-05-10
Add PR #75 review feedback for js from @stormpython @captainsafia

### LINKED ISSUE #86: Address feedback from PR #75

willingc:
- Addresses feedback on #75 

- Partially closes #86

### DIFF
--- a/ci/jupyter.js
+++ b/ci/jupyter.js
@@ -104,7 +104,6 @@ class JupyterServer {
     // rx-jupyter for writing a notebook to the contents API
     const apiPath = "/api/bookstore/published/";
     const xhr = await ajax({
-      // TODO: Resolve https://github.com/nteract/bookstore/pull/75#discussion_r280785802
       url: url_path_join(this.endpoint, apiPath, path),
       responseType: "json",
       createXHR: () => new XMLHttpRequest(),
--- a/ci/jupyter.js
+++ b/ci/jupyter.js
@@ -104,6 +104,7 @@ class JupyterServer {
     // rx-jupyter for writing a notebook to the contents API
     const apiPath = "/api/bookstore/published/";
     const xhr = await ajax({
+      // TODO: Resolve https://github.com/nteract/bookstore/pull/75#discussion_r280785802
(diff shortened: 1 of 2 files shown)


mpacer:
This is good to merge if we drop the todo, as I would like to resolve the `url_path_join` discussion in favour of leaving it there as a convenience function.

It has saved me a lot of time and avoided a lot of pain, particularly around `base_url`s vs `default_url`s &c. with the jupyter server and it's approach to configuration.

mpacer:
That said, as illustrated by #92, we aren't able to test both the GET and POST handling for cloning as of right now, only the GET.

Accordingly we didn't catch that base_urls were not being appropriately respected. I'm curious how we could accomplish this without needing to bring in the entirety of something like selenium to emulate user behaviour in this context. 

@captainsafia do you have thoughts on a good way to do this without introducing too much overhead?

ci/jupyter.js
```diff
@@ -104,6 +104,7 @@ class JupyterServer {
     // rx-jupyter for writing a notebook to the contents API
     const apiPath = "/api/bookstore/published/";
     const xhr = await ajax({
+      // TODO: Resolve https://github.com/nteract/bookstore/pull/75#discussion_r280785802
```
mpacer: In line with the reasoning behind bringing `url_path_join` into its own module for user clarity, I would suggest that having this utility here is simpler than introducing a new dependency (especially because as far as I can tell a dependency with this functionality does not exist).

```suggestion
```

willingc: I'm okay with that approach. If we find down the road that we need more, we can do so later.

willingc:
@mpacer Thanks and :shipit:

ci/jupyter.js
mpacer: I don't seem to be able to commit the suggestion to your repository, did you turn off allow edits from maintainers on this PR?

willingc:
Usually I have push off by default on PRs. I accepted the suggestion and will merge now. Thanks @mpacer

PR #89, opened 2019-05-10
Review misc files from PR #75

### LINKED ISSUE #86: Address feedback from PR #75

willingc:
- add TODO to check security XSRF from review in #75 
- Partially closes #86

### DIFF
--- a/bookstore/clone.html
+++ b/bookstore/clone.html
@@ -4,65 +4,69 @@ Copyright (c) Jupyter Development Team.
 Distributed under the terms of the Modified BSD License.
 -->
 <html lang="en">
-  <head>
-    <meta charset="utf-8" />
-    <meta name="google" content="notranslate" />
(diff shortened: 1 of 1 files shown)


mpacer:
@captainsafia did I not address your concern by creating a `getCookieByName` function as you had requested? I'm a little confused as to what needs to be addressed here…

captainsafia:
Catching up with this conversation.

I was referring to creating a getCookieByName function that did not use regex. The getCookie method used in the notebook code base is copied from the Tornado docs. While it suffices, I think an alternate approach (like splitting on semi-colons and parsing the key values) is easier to read. It's also the approach taken by the [js-cookie](https://github.com/js-cookie/js-cookie) package that we use rx-jupyter.

willingc:
Opened #104 and closing this PR in preparation for release.

PR #90, opened 2019-05-10
Add PR #75 review feedback for python files from @MSeal and @willingc

### LINKED ISSUE #86: Address feedback from PR #75

willingc:
- Addresses #75 feedback for Python files
- Partially closes #86

### DIFF
--- a/bookstore/__init__.py
+++ b/bookstore/__init__.py
@@ -1,14 +1,14 @@
-# TODO: refactor the import, PACKAGE_DIR, del
 import os
-
-from ._version import __version__, version_info
-from .archive import BookstoreContentsArchiver
-from .bookstore_config import BookstoreSettings
-from .handlers import load_jupyter_server_extension
 
 PACKAGE_DIR: str = os.path.realpath(os.path.dirname(__file__))
 
 del os
+
+from .archive import BookstoreContentsArchiver
+from .bookstore_config import BookstoreSettings
+from .handlers import load_jupyter_server_extension
+from ._version import __version__
+from ._version import version_info
 
 
 def _jupyter_server_extension_paths():
--- a/bookstore/__init__.py
+++ b/bookstore/__init__.py
@@ -1,14 +1,14 @@
+# TODO: refactor the import, PACKAGE_DIR, del
 import os
+
+from ._version import __version__, version_info
+from .archive import BookstoreContentsArchiver
+from .bookstore_config import BookstoreSettings
+from .handlers import load_jupyter_server_extension
 
 PACKAGE_DIR: str = os.path.realpath(os.path.dirname(__file__))
 
 del os
-
-from .archive import BookstoreContentsArchiver
-from .bookstore_config import BookstoreSettings
-from .handlers import load_jupyter_server_extension
-from ._version import __version__
-from ._version import version_info
 
 
 def _jupyter_server_extension_paths():
--- a/bookstore/client/nb_client.py
+++ b/bookstore/client/nb_client.py
@@ -30,7 +30,7 @@ from notebook.notebookapp import list_running_servers
 
 
 def extract_kernel_id(connection_file):
-    # TODO should this revert to not using regex ?
+    # regex is used as a more robust approach than lstrip
     connection_filename = os.path.basename(connection_file)
     kernel_id = re.sub(r"kernel-(.*)\.json", r"\1", connection_filename)
     return kernel_id
--- a/bookstore/client/nb_client.py
+++ b/bookstore/client/nb_client.py
@@ -219,4 +219,5 @@ class CurrentNotebookClient(NotebookClient):
 
     @property
     def kernel_id(self):
+        # TODO determine if extract kernel id is used more than once
         return extract_kernel_id(self.connection_file)
--- a/bookstore/client/nb_client.py
+++ b/bookstore/client/nb_client.py
@@ -136,8 +136,7 @@ class NotebookClient:
         """Current notebook sessions. Reissues request on each call.
         """
         return {
-            session['kernel']['id']: NotebookSession(**session)
-            for session in self.get_sessions()
+            session['kernel']['id']: NotebookSession(**session) for session in self.get_sessions()
         }
 
     @property
@@ -202,9 +201,7 @@ class NotebookClientCollection:
             for session_id, session in session_dict.items():
                 if session.kernel.id == current_kernel_id:
                     return next(
-                        client
-                        for client in cls.nb_client_gen()
-                        if client.url == server_url
+                        client for client in cls.nb_client_gen() if client.url == server_url
                     )
 
 
--- a/bookstore/client/nb_client.py
+++ b/bookstore/client/nb_client.py
@@ -82,21 +82,21 @@ class NotebookSession:  # (NamedTuple):
     #     id: str #'68d9c58f-c57d-4133-8b41-5ec2731b268d',
     #     path: str #'Untitled38.ipynb',
     #     name: str #'',
-    #     kind: str #'notebook',
+    #     type: str #'notebook',
     #     kernel: KernelInfo
     #     notebook: dict # deprecated API {'path': 'Untitled38.ipynb', 'name': ''}}}
 
-    def __init__(self, *args, path, name, kind, kernel, notebook, **kwargs):
+    def __init__(self, *args, path, name, type, kernel, notebook, **kwargs):
         self.model = {
             "path": path,
             "name": name,
-            "kind": kind,
+            "type": type,
             "kernel": kernel,
             "notebook": notebook,
         }
         self.path = path
         self.name = name
-        self.kind = kind
+        self.type = type
         self.kernel = KernelInfo(**kernel)
         self.notebook = notebook
 
@@ -136,7 +136,8 @@ class NotebookClient:
         """Current notebook sessions. Reissues request on each call.
         """
         return {
-            session['kernel']['id']: NotebookSession(**session) for session in self.get_sessions()
+            session['kernel']['id']: NotebookSession(**session)
+            for session in self.get_sessions()
         }
 
     @property
@@ -201,7 +202,9 @@ class NotebookClientCollection:
             for session_id, session in session_dict.items():
                 if session.kernel.id == current_kernel_id:
                     return next(
-                        client for client in cls.nb_client_gen() if client.url == server_url
+                        client
+                        for client in cls.nb_client_gen()
+                        if client.url == server_url
                     )
 
 
--- a/bookstore/client/nb_client.py
+++ b/bookstore/client/nb_client.py
@@ -18,9 +18,9 @@ Example
   'token': '',
   'url': 'http://localhost:8888/'}]
 """
+import json
 import os
 import re
-import json
 from copy import deepcopy
 from typing import NamedTuple
 
@@ -30,6 +30,7 @@ from notebook.notebookapp import list_running_servers
 
 
 def extract_kernel_id(connection_file):
+    # TODO should this revert to not using regex ?
     connection_filename = os.path.basename(connection_file)
     kernel_id = re.sub(r"kernel-(.*)\.json", r"\1", connection_filename)
     return kernel_id
@@ -81,21 +82,21 @@ class NotebookSession:  # (NamedTuple):
     #     id: str #'68d9c58f-c57d-4133-8b41-5ec2731b268d',
     #     path: str #'Untitled38.ipynb',
     #     name: str #'',
-    #     type: str #'notebook',
+    #     kind: str #'notebook',
     #     kernel: KernelInfo
     #     notebook: dict # deprecated API {'path': 'Untitled38.ipynb', 'name': ''}}}
 
-    def __init__(self, *args, path, name, type, kernel, notebook, **kwargs):
+    def __init__(self, *args, path, name, kind, kernel, notebook, **kwargs):
         self.model = {
             "path": path,
             "name": name,
-            "type": type,
+            "kind": kind,
             "kernel": kernel,
             "notebook": notebook,
         }
         self.path = path
         self.name = name
-        self.type = type
+        self.kind = kind
         self.kernel = KernelInfo(**kernel)
         self.notebook = notebook
 
--- a/bookstore/client/store_client.py
+++ b/bookstore/client/store_client.py
@@ -14,7 +14,6 @@ class BookstoreClient(CurrentNotebookClient):
 
     @property
     def publish_endpoint(self):
-        # TODO: refactor to /api/bookstore/publish/ ?
         api_endpoint = "/api/bookstore/published/"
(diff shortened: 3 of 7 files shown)


bookstore/__init__.py
```diff
@@ -1,14 +1,14 @@
+# TODO: refactor the import, PACKAGE_DIR, del
```
MSeal: Let's just remove the PACKAGE_DIR from this file. We can just move it to where it's used in `clone.py`

bookstore/client/nb_client.py
```diff
@@ -30,6 +30,7 @@
 
 
 def extract_kernel_id(connection_file):
+    # TODO should this revert to not using regex ?
```
MSeal: Yes.

bookstore/client/store_client.py
```diff
@@ -7,37 +7,38 @@
...
 
-    def __init__(self, s3_bucket=""):
+    def __init__(self, s3_bucket=None):
         if s3_bucket:
             self.default_bucket = s3_bucket
         super().__init__()
 
     @property
     def publish_endpoint(self):
+        # TODO: refactor to /api/bookstore/publish/ ?
```
MSeal: Agreed. Can be a separate PR.

bookstore/__init__.py
mpacer: I generally prefer to do this at the `__init__.py` so that it can be accessible from other locations, including those outside of the library.

Could you explain the reason you disprefer this pattern?

bookstore/client/nb_client.py
mpacer: No I disagree. The previous version was riddled with bugs. Regex is the only clean way to extract a prefix in python. `connection_filename.lstrip("kernel-") `  would strip all instances of the characters "e" "k" "r" "n" "l" and "-". So, when the connection file began with an "e" this would result in being unable to find the connection file.

bookstore/client/nb_client.py
```diff
@@ -81,21 +82,21 @@ class NotebookSession:  # (NamedTuple):
...
     #     kernel: KernelInfo
     #     notebook: dict # deprecated API {'path': 'Untitled38.ipynb', 'name': ''}}}
 
-    def __init__(self, *args, path, name, type, kernel, notebook, **kwargs):
+    def __init__(self, *args, path, name, kind, kernel, notebook, **kwargs):
         self.model = {
             "path": path,
             "name": name,
-            "type": type,
+            "kind": kind,
```
mpacer: This no longer is able to accept the literal response from the NotebookSession API… to not have this defeats the purpose of the type class. 
This has the form:
https://github.com/jupyter/notebook/blob/01b084381c367bdca1d58b176428b728a67628ca/notebook/services/api/api.yaml#L761-L778
```
  Session:
    description: A session
    type: object
    properties:
      id:
        type: string
        format: uuid
      path:
        type: string
        description: path to the session
      name:
        type: string
        description: name of the session
      type:
        type: string
        description: session type
      kernel:
        $ref: '#/definitions/Kernel'
```

bookstore/client/store_client.py
```diff
@@ -7,37 +7,38 @@
...
         api_endpoint = "/api/bookstore/published/"
         return f"{self.url}{api_endpoint}"
 
     def publish(self, path=None):
         """Publish notebook to bookstore"""
         if path is None:
             path = self.session.path
-        nb_json = self.get_contents(self.session.path)['content']
+        nb_json = self.get_contents(path)['content']
```
mpacer: thank you for catching that bug!

bookstore/client/store_client.py
mpacer: This is specifically a request from @rgbkrk to have this be more REST-ful in its naming. Rather than "publish" the action this is intended to be a reference to a location "published" that refers to the notebooks that are there.

I find it somewhat confusing as well, but I don't want to break from this approach without getting @rgbkrk's sign off.

bookstore/client/store_client.py
```diff
@@ -7,37 +7,38 @@
...
+        target_url = f"{self.publish_endpoint}{path}"
 
-        resp = self.req_session.put(target_url, json=json_body)
-        return resp
+        response = self.req_session.put(target_url, json=json_body)
+        return response
 
     @property
     def clone_endpoint(self):
+        # TODO: refactor to /api/bookstore/clone/ ?
```
mpacer: Same comments as above around "published"… this is specifically a request from @rgbkrk to have this be more REST-ful in its naming. Rather than "clone" the action, this is intended to be a reference to a location "cloned" that refers to the notebooks that are there.

I find this even more confusing than the `/published` name, but, again, I don't want to break from this approach without getting @rgbkrk's sign off.

bookstore/clone.py
```diff
@@ -24,24 +22,24 @@ def initialize(self):
...
         self.session = aiobotocore.get_session()
 
-    async def _clone(self, s3_bucket, file_key):
-        path = file_key
+    async def _clone(self, s3_bucket, s3_key_file):
 
         self.log.info(f"bucket: {s3_bucket}")
-        self.log.info(f"key: {file_key}")
-        full_s3_path = s3_path(s3_bucket, path)
+        self.log.info(f"key: {s3_key_file}")
```
mpacer: I think this should probably be `s3_file_key` or `s3_object_key` rather than `s3_key_file`.

bookstore/publish.py
```diff
@@ -1,16 +1,12 @@
...
-from notebook.base.handlers import path_regex
+from notebook.base.handlers import APIHandler, path_regex
 from notebook.utils import url_path_join
 from tornado import web
 
-from .s3_paths import s3_path
-from .s3_paths import s3_key
-from .s3_paths import s3_display_path
 from .bookstore_config import BookstoreSettings
+from .s3_paths import s3_display_path, s3_key, s3_path
```
mpacer: I had split these out because of your previous styling edit that took my one line imports and split them into multiple. I prefer them to be on one line, but was trying to incorporate your changes.

Can you give me guidance on when to go in one direction or the other?

ci/clone_request.py
```diff
@@ -33,24 +35,11 @@ def post(**kwargs):
...
-
-    # tests minio created 'bookstore' bucket
-    # res = get("?s3_bucket=bookstore&s3_key=ci-published/ci-published.ipynb")
-    # print(res.content)
-    # res = post(s3_bucket="nteract-notebooks", s3_key="Introduction_to_Chainer.ipynb")
-    # print(res.content)
-    res = post(s3_bucket="bookstore", s3_key="ci-published/ci-published.ipynb")
-    pprint.pprint(res.json())
-    res = post(
+    response = post(s3_bucket="bookstore", s3_key="ci-published/ci-published.ipynb")
```
mpacer: The comments are purposefully there to illustrate other things that are intended to be easily tested interactively. They should not be removed.

mpacer: Or rather…if we are going to remove these comments, we might as well remove the file in its entirety as it's role is in aiding debugging. I thought including it in the repo would be useful for others to have as an example. 

To be honest, the motivation of the client was to make it easier to do this kind of quick debugging.

mpacer:
@MSeal re: your comments on testing. 

I would love your insight on how to test the various clients. I am super enthusiastic about that just had a tricky time getting that to start as to do it properly requires some pretty extensive integration test work. 

Regarding the cloning functionality itself that was introduced in #75, that was what the js based integration tests were testing.

ci/clone_request.py
willingc: Honestly, I thought this was uncommented code not a comment for the user of the library. It would be better to perhaps put in a docstring instead and make it clear that these are tips for debugging.

bookstore/client/nb_client.py
MSeal: Then maybe we should leave "type" as a key but make the api function signature rename the argument variable. Having a reserved word variable name is problematic in python.

bookstore/publish.py
MSeal: I tend to let the linter do whatever the linter likes unless it's really jarring. I generally leave all import from the same path in the same statement and let it spill to multi-line over separate repeated imports. Otherwise it's easy to miss duplicate imports and adds confusion around what importing is doing.Though that's kjust my opinion on the topic.

bookstore/clone.py
MSeal: It's odd that we force keys to be passed around rather than keeping credential collections purely within the s3 client. This makes setups where there's no key file unable to use this tool despite having valid access to S3.

MSeal:
@mpacer I would argue we need unittests first here most of all. You'll need to mock s3 and web interfaces above and below the various functions, but there's a lot of logic here for handling edge cases that need coverage for reliable development. A starting place would be to get a test for clone `get` receiving and empty file path. I can help set up first first couple tests when I get back if it'd help get the ball rolling.

bookstore/client/nb_client.py
mpacer: The function signature is what needs to have it. I'd be fine making it an attribute that has a different name (it's less ideal & goes against the goals of the API… but possible), but being able to unpack the json response without altering it is the specific intent of this API. 

Also, it's technically not a reserved keyword… just a super commonly used function that is not best practice to overwrite. Keywords: https://docs.python.org/3/reference/lexical_analysis.html#keywords. Otherwise this would be a syntax error.

I don't like doing this either and if it weren't for the explicit desire to be isomorphic to the response object I wouldn't push on this.

bookstore/clone.py
mpacer: these aren't s3_keys. That's what I'm saying. They are paths to the object we should come up with a better name for them as they clearly just lead to confusion.

bookstore/publish.py
mpacer: I agree, I prefer to let the linter decide, but I was trying to incorporate the changes @willingc  had made in this PR: https://github.com/nteract/bookstore/pull/70/files#diff-b79c22e74e3095962bbd282b0bd0c040R13

ci/clone_request.py
mpacer: If their purpose is that ambiguous shouldn't we just remove them from the library?

bookstore/clone.py
mpacer: Ah perhaps we should call it s3_object_key? that's the language they use in the s3 docs: https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html#object-keys

willingc: I could get behind `s3_object_key_name` which seems to accurately reflect the AWS docs and not imply that we are passing around keys or files.

bookstore/client/store_client.py
MSeal: Let's change it, there was general consensus from folks that the action routes shouldn't be past-tensed, but retrieval routes can be nouns. If it's generating the object let's leave it a present tense, it it's grabbing older generated objects then the past tense or pluralized nouns is probably correct.

bookstore/client/nb_client.py
MSeal: Just wanted to raise it as it can cause unexpected issues in the future. Not a blocker or required change.

bookstore/clone.py
mpacer: I went with `s3_object_key` as that is the way that it is referred to most generally in the docs and is a slight bit shorter. Should we change the url param to match? Since it hasn't been officially released yet, now is probably the time to do it.

bookstore/client/store_client.py
mpacer: Ok, a change PR is underway.

bookstore/clone.py
willingc: Yep. I would make it consistent. Thanks @mpacer

bookstore/client/store_client.py
rgbkrk: It's probably ok to just refer to it in an RPC way, especially since I'm the only one who thought otherwise. I'd prefer what makes sense to most people.

willingc: I think at this point that we should leave as is.

bookstore/__init__.py
willingc: I have removed the TODO to move this forward. We can always refactor at a later time if needed.

ci/clone_request.py
willingc: I've updated the docstrings so that the content remains and is more understandable.

bookstore/clone.py
willingc: I changed it in this PR.

willingc:
@mpacer I spent some time this morning going through this PR in detail. All tests are passing (thanks to M's fix to the js ci - changing const to let). Let's merge this and move forward.

We're getting very close to a release. Just a few more tests and a changelog after this lands.

ci/utils.js
```diff
@@ -9,7 +9,7 @@ const url_path_join = function(...pieces) {
   // """
   const initial = pieces[0].startsWith("/");
   const final = pieces[pieces.length - 1].endsWith("/");
-  const result = pieces
+  let result = pieces
```
mpacer: I am a little confused — shouldn't this already be changed on master?

https://github.com/nteract/bookstore/blob/ad44c3abad19dce0a834e320cd30a5b18f2ccc29/ci/utils.js#L10-L15

willingc: It is but this predated that so it is probably a merge conflict now that needs to be taken care of in this PR so that master is used.

bookstore/client/nb_client.py
```diff
@@ -81,21 +82,21 @@ class NotebookSession:  # (NamedTuple):
     #     id: str #'68d9c58f-c57d-4133-8b41-5ec2731b268d',
     #     path: str #'Untitled38.ipynb',
     #     name: str #'',
-    #     type: str #'notebook',
+    #     kind: str #'notebook',
     #     kernel: KernelInfo
     #     notebook: dict # deprecated API {'path': 'Untitled38.ipynb', 'name': ''}}}
 
-    def __init__(self, *args, path, name, type, kernel, notebook, **kwargs):
+    def __init__(self, *args, path, name, kind, kernel, notebook, **kwargs):
```
mpacer: I would prefer that we don't change this as it would make it so that we cannot directly consume the contents of the API response from the `notebook`'s `/api/sessions/` in https://github.com/nteract/bookstore/pull/90/files#diff-7d77120266750554a7e864156f790108R139

bookstore/client/nb_client.py
```diff
@@ -30,6 +30,7 @@
 
 
 def extract_kernel_id(connection_file):
+    # regex is used as a more robust approach than lstrip
```
mpacer: Many thanks for the clarifying comment! Looks great :D!

ci/utils.js
willingc: Rebased

willingc:
@mpacer I think that this has all the changes that we discussed on our call.

PR #91, opened 2019-05-11
Add version check and tests

willingc:


bookstore/_version.py
```diff
@@ -21,3 +21,23 @@
 version_info = (2, 3, 0, "", "dev")
 
 __version__ = ".".join(map(str, version_info[:3])) + ".".join(version_info[3:])
+
+
+def _check_version(bookstore_version, log):
```
MSeal: Maybe give the function a boolean return value that indicates if the request version is supported or not?

bookstore/_version.py
```diff
@@ -21,3 +21,23 @@
...
+    """Check version and log status"""
+    if not bookstore_version:
+        log.warning(
+            f"Bookstore has no version header, which means it is likely < 2.0. Expected {__version__}"
+        )
+        return
```
MSeal: Don't think you need any of the returns if we're not returning a palyoad and all branches are gated on conditionals.

mpacer:
Is this primarily to get us bootstrapped into having more tests in bookstore? Otherwise, could you say where we expect the version check to be used outside of the tests?

Note: if we were going to implement something pep440 checking (like what we have in nbconvert)… I could definitely see the criticality for having this kind of version number check. 

for reference nbconvert has a utility `create_valid_version`  https://github.com/jupyter/nbconvert/blob/06f3a90f6b7b578a78efe1f17138a4fea43fcfa5/nbconvert/_version.py#L5-L60

```python
def create_valid_version(release_info, epoch=None, pre_input='', dev_input=''):
    '''
    Creates a pep440 valid version of version number given a tuple integers
    and optional epoch, prerelease and developmental info.
    Parameters
    ----------
    release_info : Tuple(Int)
    epoch : Int, default None
    pre_input : Str, default ''
    dev_input : Str, default ''
    '''
```

willingc:
This was mostly to get us started. We can either close or merge. No strong preference either way.

PR #92, opened 2019-05-15
Address base_urls properly; handle path creation on the back end

mpacer:
Currently, the clone landing page from the GET api does not properly resolve base_urls. This addresses it. 

As it relates to https://github.com/nteract/bookstore/pull/75#discussion_r280785802 and the todo in https://github.com/nteract/bookstore/pull/88:
This is a case where because we lac a `url_path_join` utility in the front-end, it is easier to rely on the one provided by the notebook package from the backend rather than doing the text parsing for ourselves.

For example, working with `base_url`s causes a lot of headaches.

```
ipdb> self.default_url
'/blah/tree'
ipdb> self.base_url
'/blah/'
```

@willingc given your experience from JupyterHub & it's routing & base_url paradigms is there a better way to get access to the host name & protocol than to pull it from the request?

### DIFF
--- a/bookstore/utils.py
+++ b/bookstore/utils.py
@@ -1,0 +1,1 @@
+from notebook.utils import url_path_join
--- a/bookstore/utils.py
+++ b/bookstore/utils.py
@@ -1,1 +1,16 @@
-from notebook.utils import url_path_join
+def url_path_join(*pieces):
+    """Join components of url into a relative url
+    Use to prevent double slash when joining subpath. This will leave the
+    initial and final / in place
+    """
+    initial = pieces[0].startswith('/')
+    final = pieces[-1].endswith('/')
+    stripped = [s.strip('/') for s in pieces]
+    result = '/'.join(s for s in stripped if s)
+    if initial:
+        result = '/' + result
+    if final:
+        result = result + '/'
+    if result == '//':
+        result = '/'
+    return result
--- a/bookstore/utils.py
+++ b/bookstore/utils.py
@@ -1,7 +1,13 @@
+"""Utility and helper functions."""
+
+
 def url_path_join(*pieces):
-    """Join components of url into a relative url
+    """Join components into a relative url.
+    
     Use to prevent double slash when joining subpath. This will leave the
-    initial and final / in place
+    initial and final / in place.
+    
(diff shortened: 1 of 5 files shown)


willingc:
Instead of importing `url_path_join` from the notebook, let's put a `url_path_join` into a `utils.py` file here. You can model the initial implementation on the code in the notebook repo. 

We've done the same thing in jupyterhub. This prevents having to rely on an implementation that may change in the notebook as well as reducing the cognitive load for a new developer who will wonder why the notebook import is needed.

mpacer:
@willingc Awesome, I'm happy to do that! I had thought the intention suggested by @captainsafia was to rely on external libraries when possible for this functionality. 

Those are both great reasons to do this! Thank you, for the suggestions. I now know what to look for going forward as reasons to pursue these decisions.

bookstore/utils.py
```diff
@@ -0,0 +1,16 @@
+def url_path_join(*pieces):
+    """Join components of url into a relative url
```
willingc:
```suggestion
    """Join components into a relative url.
    
```

bookstore/utils.py
```diff
@@ -0,0 +1,16 @@
...
+    Use to prevent double slash when joining subpath. This will leave the
+    initial and final / in place
```
willingc:
```suggestion
    initial and final / in place.
    
    Code based on Jupyter notebook `url_path_join`.
```

bookstore/utils.py
```diff
@@ -0,0 +1,16 @@
+def url_path_join(*pieces):
```
willingc:
```suggestion
"""Utility and helper functions."""


def url_path_join(*pieces):
```

mpacer:
Those suggestions look great & have been committed! Many thanks :D!

mpacer:
@willingc since you and @MSeal both approved, is it still violating the no self-merge policy for me to merge this?

mpacer:
Also, thank you for merging it! 💥

willingc:
@mpacer Nope. If you have addressed the feedback and folks have approved, it's fine to self merge (unless there are a bunch of complex code changes). 👍

Issue #95, opened 2019-05-16
Document public API


willingc:
Using markdown or swagger, we should have a one page document of the public facing API.

mpacer:
Closed by #98

Issue #96, opened 2019-05-20
Prepare for 2.3.0 Release


willingc:
Some tasks to do before 2.3.0 release:

- [x] Add Python unit tests (aim for 50% coverage for this release and increase in following releases)
- [x] Add an API document #95 
- [x] PR #89 Misc fixes
- [x] PR #90 Python files
- [x] PR #91 Version check test. Assigned to @willingc
- [x] PR #88 JS fixes
- [x] Update changelog

cc/ @mpacer @MSeal @rgbkrk

MSeal:
@mpacer and I got started on the unit testing front. M's going to finish that up but we should have a better setup now for testing the async handlers moving forward.

willingc:
Let's work on the changelog and then :shipit:

mpacer:
With the three PRs that are open now it should bring the overall code-base to 84% test coverage.

More importantly, it's far better documented and way more readable.

I'll work on the changelog assuming that those 3 PRs will be merged. It changes a few of the APIs, but fortunately we now have API docs so its clear what exactly changed! Huzzah!

mpacer:
Released! 

Closing thanks all!

PR #98, opened 2019-05-24
Draft api docs

mpacer:
I didn't include anything about the `ArchiveContentsManager`, because it doesn't really have an API in the sense of the handlers.

This should address one of #96's points.

- [x] clone APIs
- [x] publish API
- [x] version API


mpacer:
Closes #95

willingc:
Looks great and am merging as is. We can refine wording over time if needed. :tada:

PR #100, opened 2019-05-24
Edit usage documentation

willingc:
- Make the doc more user focused
- Add a bit more text for understanding
- Image can be replaced at a later date.

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -43,8 +43,8 @@ this, bookstore allows users to set workspace and published storage paths. For e
 - ``/workspace`` - where users edit and store notebooks
 - ``/published`` - notebooks to be shared to an organization
 
-Bookstore archives versions by keeping the path intact (until a user changes them). For example, the prefixes that could
-be associated with storage types:
+Bookstore archives notebook versions by keeping the path intact (until a user changes them). For example, the prefixes
+that could be associated with storage types:
 
 - Notebook in "draft" form: ``/workspace/kylek/notebooks/mine.ipynb``
 - Most recent published copy of a notebook: ``/published/kylek/notebooks/mine.ipynb``
@@ -63,7 +63,7 @@ How to Publish a notebook
 -------------------------
 
 To publish a notebook, Bookstore uses a publishing endpoint which is a ``serverextension`` to the classic Jupyter
-server. If you wish to publish notebook, explicitly enable bookstore as a server extension to use the endpoint. By
+server. If you wish to publish notebooks, explicitly enable bookstore as a server extension to use the endpoint. By
 default, publishing is not enabled.
 
 To enable the extension globally, run::
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -15,11 +15,8 @@ now request different versions from the remote storage, such as S3, and
 
     User saves to Local System ------------------> Remote Data Store (i.e. S3)
 
-<<<<<<< HEAD
+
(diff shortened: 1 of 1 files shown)


docs/source/usage.rst
```diff
@@ -1,47 +1,75 @@
...
 
-Every *save* of a notebook creates an *immutable copy* of the notebook on object storage.
+                                             *store*
 
-To simplify implementation, we currently rely on S3 as the object store, using
+Data scientist saves to Local System ------------------> Remote Data Store (i.e. S3) Workspace
+
+                                             *clone*
+
+User requests a notebook to use locally <-------------- Remote Data Store (i.e. S3) Workspace
```
mpacer: Tiny thing, just to be consistent, we should probably use Workspace to mean the semantics associated with workspace. Cloning can be done from any s3 path.

```suggestion
User requests a notebook to use locally <-------------- Remote Data Store (i.e. S3)
```

docs/source/usage.rst
```diff
@@ -1,47 +1,75 @@
...
+
+How to store and clone versions
+-------------------------------
+
+Bookstore uses automatic notebook version management and specific storage paths when storing a notebook.
+
+Automatic Notebook Version Management
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Every *save* of a notebook creates an *immutable copy* of the notebook on object storage. Initially, Bookstore
```
mpacer: While it is an immutable copy, I feel like we might want to be clearer that it is overwriting the file in place. It relies on versioned s3 buckets to handle the versioning.

docs/source/usage.rst
```diff
@@ -1,47 +1,75 @@
...
-storage types:
+Bookstore archives notebook versions by keeping the path intact (until a user changes them). For example, the prefixes
+that could be associated with storage types:
 
 - Notebook in "draft" form: ``/workspace/kylek/notebooks/mine.ipynb``
 - Most recent published copy of a notebook: ``/published/kylek/notebooks/mine.ipynb``
 
-Scheduled notebooks will also be referred to by the notebook ``key``. In addition, we'll need to be able to surface
-version IDs as well.
+When scheduling execution of notebooks, each notebook path is a namespace that an external service can access. This
```
mpacer: I actually wonder if we should remove this until we support some kind of hook into a scheduling system.

docs/source/usage.rst
willingc: No problem making the change. I want to make sure that I understand correctly how we want to use the word workspace.

My understanding: workspace is a path location to an S3 store where a user stores their notebooks.

docs/source/usage.rst
willingc: Good point on the overwrite. Will add.

docs/source/usage.rst
willingc: We can either 1) remove or 2) place in an information box that says that the current release does not support but we expect to do so in a later release. Let me know which you prefer. I don't have a strong feeling either way.

docs/source/usage.rst
willingc: Added to the following paragraph.

docs/source/usage.rst
willingc: Added as a note.

rgbkrk:
Looks like review feedback is addressed. This is some additive improvement to the documentation; thank you Carol.

PR #103, opened 2019-05-29
Adds autodoc for clone and expands docstrings

willingc:


mpacer:
Looks great!

bookstore/clone.py
```diff
@@ -1,25 +1,47 @@
...
 BOOKSTORE_FILE_LOADER = FileSystemLoader(PACKAGE_DIR)
 
 
 class BookstoreCloneHandler(IPythonHandler):
-    """Clone a notebook from s3."""
+    """Handle notebook clone from storage.
+
+    Provides API handling for ``GET`` and ``POST`` when cloning a notebook
+    from storage (S3). Launches a user interface cloning options page when
+    ``GET`` is sent.
```
rgbkrk: I realize I'm late to the game here... What was the reasoning behind putting the HTML endpoint on the `GET` of `/api/bookstore/cloned`? Usually, that would belong somewhere like `/bookstore/clone`, while the `GET` on the API endpoint would either be disabled or have a JSON payload like the rest of the APIs.

willingc: I think @mpacer thought that was what you wished to do while you were on vacation. Do you want to open a PR that adds the correct naming? I think that would help us.

mpacer: If I understand correctly, Kyle isn’t disagreeing about /api/bookstore/cloned —> /api/bookstore/clone

But rather is suggesting the clone handler itself should be separated into an api handler (that does the cloning) at /api/bookstore/cloned

And a HTML handler at /bookstore/clone.

mpacer: Also I think we all resolved to change it from cloned/published to clone/publish, I just have focused on the testing rather than the endpoint change PRs since that seemed more urgent.

willingc: Let's do this. Let's review the OpenAPI yaml spec and make that the source of truth to minimize confusion (at least mine). 

If we are not using the yaml file for testing in the bookstore directory, let's remove it and keep the file in docs/source/reference. The autodoc extension for openapi will not work if the spec is not in the docs tree.

rgbkrk: Yeah, I just want to make sure that API endpoints all live on `/api/*` and HTML endpoints are on other endpoints. Beyond the need to obey REST API conventions for future developers to understand the code base and system easily, it's also important because of how the handler classes are set up for API vs non-API, including the headers that tornado sends back (and how the browser interprets them).

Sorry to add this to a doc PR, this PR just made it really clear for me to see what was happening here (so docs are working!) 😄

PR #105, opened 2019-05-30
Generate Sphinx REST API page

willingc:
- Use sphinxcontrib.openapi extension
- copies yaml spec into docs folder (wasn't sure if the spec was being used for testing so didn't move)

[Rendered API](https://test-bookstore.readthedocs.io/en/openapi-sphinx/openapi.html)


willingc:
Looks like my rebase left a stray file.

willingc:
CircleCI failure is fixed by @mpacer in #106

PR #102, opened 2019-05-29
Clone GET Testing

mpacer:
This adds testing functionality and mocking for the clone handler's GET methods. 

This mostly was focused on figuring out how to mock up most of what we need to test the handler without actually requiring a running server. 

This also factors out the parameterization for the template so that we can test that separately. 

To get the POST tested will require adding moto (to mock up s3 and interactions with `aiobotocore`).

Many thanks to @mseal for help getting the initial framework up and running.

PR #106, opened 2019-05-30
Add POST handler tests for cloning

mpacer:
This continues #102 to add more testing for the post handler.

I have not yet gotten the mock to work correctly to test the newly introduced helper function.

This also adds a quick fix to the js files that had broken the CI tests when run locally. It's concerning to me that that didn't come up in our CI testing, since updating a const should always cause an issue which suggests the tests weren't being run

### DIFF
--- a/bookstore/clone.py
+++ b/bookstore/clone.py
@@ -2,6 +2,8 @@
 import json
 import os
 
+from copy import deepcopy
+
 import aiobotocore
 from jinja2 import FileSystemLoader
 from notebook.base.handlers import IPythonHandler, APIHandler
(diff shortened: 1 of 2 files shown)


mpacer:
@MSeal Is the credentials stuff the kind of thing I should try to mock with moto?

@willingc should we choose to test the python on only one of circleci or travis? Is there an advantage to testing it in both places?

willingc:
>@willingc should we choose to test the python on only one of circleci or travis? Is there an advantage to testing it in both places?

For now, let's keep both. They are a good check against each other while we develop the tests.

willingc:
@mpacer Can we split out the quick fix for the CI from this PR? This way we have valid js tests again and we can iterate on the post tests.

willingc:
@mpacer Can you rebase this on the latest master? I'm running the python tests locally and they are passing where Travis has a couple failing. 🤷‍♀

mpacer:
> @mpacer Can you rebase this on the latest master? I'm running the python tests locally and they are passing where Travis has a couple failing. 🤷‍♀

It was already rebased on the latest master — this has to do with how s3 configuration is set up. My guess is you have an `~/.aws/credentials` file and so this isn't failing in the same way :/

this is why I think we need to use moto to mock the underlying call.

mpacer:
> @mpacer Can we split out the quick fix for the CI from this PR? This way we have valid js tests again and we can iterate on the post tests.

Done with #107

willingc:
Let's do this..rebase this PR. If the tests don't pass, remove the two tests where the mocking is not working fully. Then move those two tests to a new PR and we can merge this.

mpacer:
I think at this point I might want to rebase it on #121 since those tests will need to be on their own separate handler anyway, otherwise we're going to run into a conflict with that one where we're trying to test functions that no longer exist.

mpacer:
Well… that is a disappointing code_coverage report. I suppose that's mostly because of the refactor & the fact that without figuring out the moto stuff the tests going to exit pretty early in the method.

mpacer:
I don't know why the commits are appearing out of order, but the add mock_s3 commit is supposed to come before the other two, which is why it seems to be testing not the latest commit… something weird is happening there… not really sure.

mpacer:
They're still out of order, but I think this should improve the coverage.

mpacer:
Does anyone else know why CodeCov isn't picking up the latest version of the changes that a bunch more of the publish handler?

willingc:
I think that I configured codecov to only do master. Let's land this. We can fiddle with the dials later.

PR #107, opened 2019-05-30
Fix bug in trying to redefine const, switch to let

mpacer:
This pulls out the fix in #106 to fix CI testing for now while the POST stuff is still in the works.


willingc:
Thanks @mpacer

PR #108, opened 2019-05-31
Add tox file

willingc:
Settings shamelessly taken from papermill.


rgbkrk:
Looks like some pretty standard CI bits so I feel comfortable merging, thanks @willingc!

Issue #112, opened 2019-06-03
Tox & Conda instructions for running tests


mpacer:
There have historically been issues between tox and conda playing nicely together. We should warn contributors who use `conda` about this.

willingc:
Closed by #117

willingc:
[Link to documentation](https://bookstore.readthedocs.io/en/latest/project/python_tests.html#run-python-tests)

PR #117, opened 2019-06-03
Configure travis to use tox

### LINKED ISSUE #116: Configure Travis to use tox
Now that we have tox working locally, we should have travis use tox too.

willingc:
Closes #116 #112 

- update tox.ini for black and mypy
- add tox to travis config

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,2 @@
+[run]
+source = bookstore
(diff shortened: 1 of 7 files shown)


willingc:
I think the codecov comment above may be because this is the first use of codecov in this repo. Codecov is definitely getting information: https://codecov.io/gh/nteract/bookstore/tree/master/bookstore

PR #121, opened 2019-06-07
Refactor Clone handler

### LINKED ISSUE #109: Refactor handler & introduce /bookstore/clone endpoint
Based on comments with @rgbkrk in #105 and #103 it became clear that we need to refactor the Clone handler into an `APIHandler` and an HTML, user-facing `IPythonHandler`.

This would create a new route at `/bookstore/clone`.

@rgbkrk @willingc

mpacer:
Closes #109 

This does not address the switch in all of the APIs as well, I will add that to this PR as it seems thematically related enough to go with it (and it would run into merge conflicts with this if I were to make it an independent PR).

This introduces BookstoreCloneAPIHandler, that inherits from APIHandler rather than IPythonHandler.

### DIFF
--- a/bookstore/clone.py
+++ b/bookstore/clone.py
@@ -18,9 +18,8 @@ BOOKSTORE_FILE_LOADER = FileSystemLoader(PACKAGE_DIR)
 class BookstoreCloneHandler(IPythonHandler):
     """Handle notebook clone from storage.
 
-    Provides API handling for ``GET`` and ``POST`` when cloning a notebook
-    from storage (S3). Launches a user interface cloning options page when
-    ``GET`` is sent.
+    Provides handling for ``GET`` requests when cloning a notebook
+    from storage (S3). Launches a user interface with cloning options.
 
     Methods
     -------
@@ -30,8 +29,6 @@ class BookstoreCloneHandler(IPythonHandler):
         Checks for valid storage settings and render a UI for clone options.
     construct_template_params(self, s3_bucket, s3_object_key)
         Helper to populate Jinja template for cloning option page.
-    post(self)
-        Clone a notebook from the location specified by the payload.
     get_template(self, name)
         Loads a Jinja template and its related settings.
 
@@ -46,10 +43,10 @@ class BookstoreCloneHandler(IPythonHandler):
 
     @web.authenticated
     async def get(self):
-        """GET /api/bookstore/cloned
+        """GET /bookstore/clone?s3_bucket=<your_s3_bucket>&s3_key=<your_s3_key>
 
         Renders an options page that will allow you to clone a notebook
-        from a specific bucket.
+        from a specific bucket via the Bookstore cloning API.
         """
         s3_bucket = self.get_argument("s3_bucket")
         if s3_bucket == '' or s3_bucket == "/":
@@ -87,22 +84,15 @@ class BookstoreCloneHandler(IPythonHandler):
 class BookstoreCloneAPIHandler(APIHandler):
     """Handle notebook clone from storage.
 
-    Provides API handling for ``GET`` and ``POST`` when cloning a notebook
-    from storage (S3). Launches a user interface cloning options page when
-    ``GET`` is sent.
+    Provides API handling for ``POST`` when cloning a notebook
+    from storage (S3).
 
     Methods
     -------
     initialize(self)
         Helper to access bookstore settings.
-    get(self)
-        Checks for valid storage settings and render a UI for clone options.
-    construct_template_params(self, s3_bucket, s3_object_key)
-        Helper to populate Jinja template for cloning option page.
     post(self)
         Clone a notebook from the location specified by the payload.
-    get_template(self, name)
-        Loads a Jinja template and its related settings.
 
     See also
     --------
--- a/bookstore/clone.py
+++ b/bookstore/clone.py
@@ -4,7 +4,7 @@ import os
 
 import aiobotocore
 from jinja2 import FileSystemLoader
-from notebook.base.handlers import IPythonHandler
+from notebook.base.handlers import IPythonHandler, APIHandler
 from tornado import web
 
 from . import PACKAGE_DIR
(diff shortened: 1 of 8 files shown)


mpacer:
With 7b46868 this Closes #111

mpacer:
This is odd, I forgot to update all the URIs in the test_clone.py tests, but they didn't break. 

@MSeal That seems concerning, weren't these supposed to be checking that the APIs remained constant? 

In particular, I'd have thought `test_get_success` would fail. Any thoughts as to why it didn't?

mpacer:
Hrm… this seems to be related to the way that we're mocking the Tornado Application… I think because we're not using any of it's routing logic, this actually doesn't test the APIs themselves, only how the handlers handle arguments

bookstore/clone.py
```diff
@@ -18,9 +18,8 @@
 class BookstoreCloneHandler(IPythonHandler):
     """Handle notebook clone from storage.
 
-    Provides API handling for ``GET`` and ``POST`` when cloning a notebook
-    from storage (S3). Launches a user interface cloning options page when
-    ``GET`` is sent.
+    Provides handling for ``GET`` requests when cloning a notebook
```
willingc: Nice!

bookstore/clone.py
```diff
@@ -40,6 +37,68 @@ class BookstoreCloneHandler(IPythonHandler):
...
+        """Helper to retrieve bookstore setting for the session."""
+        self.bookstore_settings = BookstoreSettings(config=self.config)
+
+    @web.authenticated
+    async def get(self):
+        """GET /bookstore/clone?s3_bucket=<your_s3_bucket>&s3_key=<your_s3_key>
+
+        Renders an options page that will allow you to clone a notebook
+        from a specific bucket via the Bookstore cloning API.
+        """
```
willingc: Let's add a note in this docstring about the distinction between s3_object_key and s3_key. In particular let's clarify which that the key passed as a parameter in the example above is the s3_key (not the s3_object_key)

bookstore/clone.py
```diff
@@ -40,6 +37,68 @@ class BookstoreCloneHandler(IPythonHandler):
...
+        s3_bucket = self.get_argument("s3_bucket")
+        if s3_bucket == '' or s3_bucket == "/":
+            raise web.HTTPError(400, "Must have a bucket to clone from")
+
+        # s3_paths module has an s3_key function; s3_object_key avoids confusion
+        s3_object_key = self.get_argument("s3_key")
+        if s3_object_key == '' or s3_object_key == '/':
+            raise web.HTTPError(400, "Must have a key to clone from")
```
willingc:
```suggestion
            raise web.HTTPError(400, "Requires an S3 key")
```

bookstore/clone.py
```diff
@@ -40,6 +37,68 @@ class BookstoreCloneHandler(IPythonHandler):
...
+    @web.authenticated
+    async def get(self):
+        """GET /bookstore/clone?s3_bucket=<your_s3_bucket>&s3_key=<your_s3_key>
+
+        Renders an options page that will allow you to clone a notebook
+        from a specific bucket via the Bookstore cloning API.
+        """
+        s3_bucket = self.get_argument("s3_bucket")
+        if s3_bucket == '' or s3_bucket == "/":
+            raise web.HTTPError(400, "Must have a bucket to clone from")
```
willingc:
```suggestion
            raise web.HTTPError(400, "Requires an S3 bucket")
```

bookstore/clone.py
```diff
@@ -40,6 +37,68 @@ class BookstoreCloneHandler(IPythonHandler):
...
+            raise web.HTTPError(400, "Must have a key to clone from")
+
+        self.log.info("Setting up cloning landing page from %s", s3_object_key)
+
+        template_params = self.construct_template_params(s3_bucket, s3_object_key)
+        self.set_header('Content-Type', 'text/html')
+        self.write(self.render_template('clone.html', **template_params))
+
+    def construct_template_params(self, s3_bucket, s3_object_key):
+        """Helper that takes valid S3 parameters and populates UI template"""
```
willingc:
```suggestion
        """Helper that takes valid S3 parameters and populates UI template
        
        Returns
        --------
        
        dict
            Template parameters in a dictionary
        """
```

bookstore/clone.py
```diff
@@ -40,6 +37,68 @@ class BookstoreCloneHandler(IPythonHandler):
...
+
+    def get_template(self, name):
+        """Loads a Jinja template by name."""
+        return BOOKSTORE_FILE_LOADER.load(self.settings['jinja2_env'], name)
+
+
+class BookstoreCloneAPIHandler(APIHandler):
+    """Handle notebook clone from storage.
+
+    Provides API handling for ``POST`` when cloning a notebook
```
willingc: Perhaps

```suggestion
    Provides API handling for ``POST`` and clones a notebook
```

bookstore/clone.py
willingc: Let's change line 19 to: Prepares a clone options page and populates the UI with clone option parameters.

bookstore/clone.py
```diff
@@ -40,6 +37,68 @@ class BookstoreCloneHandler(IPythonHandler):
...
+        self.log.info("Setting up cloning landing page from %s", s3_object_key)
+
+        template_params = self.construct_template_params(s3_bucket, s3_object_key)
+        self.set_header('Content-Type', 'text/html')
+        self.write(self.render_template('clone.html', **template_params))
+
+    def construct_template_params(self, s3_bucket, s3_object_key):
+        """Helper that takes valid S3 parameters and populates UI template"""
+        base_uri = f"{self.request.protocol}://{self.request.host}"
+        clone_api_url = url_path_join(base_uri, self.base_url, "/api/bookstore/cloned")
```
willingc: Should this end with clone?

mpacer: Indeed!

bookstore/clone.py
mpacer: I think I'd like to indicate that this is specifically an issue around cloning, because it does not require the s3_bucket to be set from the settings, and so should be distinguished from the issues that arise around working without an s3_bucket on the archiving/publishing side.

bookstore/clone.py
mpacer: I feel like conveying that this is functionality related to cloning would also be useful here, what is the reason for removing that info? Other than the loss of the cloning specific info, the rephrasing improves clarity

bookstore/clone.py
mpacer: Sounds good — I'll have a docstring & error string changing commit en-route to also incorporate your improved phrasing of "Requires" around the errors below.

bookstore/clone.py
willingc: How about my suggestion with `for clone` or `to clone from`

bookstore/clone.py
willingc: Same as above. Add clone to the end of the message.

bookstore/clone.py
mpacer: I think this comment might be coming from the same confusion as earlier. 
The `s3_key` in the above as a parameter is the `s3_object_key` internal to the function. 

```python
        s3_object_key = self.get_argument("s3_key")
        if s3_object_key == '' or s3_object_key == '/':
            raise web.HTTPError(400, "Must have a key to clone from")
```

This is an attempt to fit with convention that most S3 endpoints & clients have where what is provided is only the "Bucket" and the "Key", and no authentication or authorization ever happens over the web. In those contexts
`Key`, `s3Key`, `s3_key`, `S3Key` is the parameter that we landed on calling `s3_object_key` internal to the function. 

The stuff around authentication and authorization are usually passed as command line args, via configuration files, or environment variables (from https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html)
AWS_ACCESS_KEY_ID – Specifies an AWS access key associated with an IAM user or role. AWS_SECRET_ACCESS_KEY – Specifies the secret key associated with the access key. This is essentially the "password" for the access key.

I think these might be unsafe to pass via URL, which it's possible that so much consolidation around the meaning of `Key` in endpoints/clients being the "object key".

If you think we should change the URL parameter to `s3_object_key` I'm on board… as it would be clearer for a user who is unfamiliar with s3 clients. We will then be deviating further from the norm in how most s3 clients & endpoints backed by s3 clients. And so we should be aware we are making that tradeoff.

bookstore/clone.py
mpacer: I just pushed changes for "in order to clone" I hadn't seen your response yet because in between writing the change and pushing it i was writing the long response to your question about s3_key

mpacer: Sorry about that!

bookstore/clone.py
mpacer: It also should have began with bookstore! yay for tests actually failing!

bookstore/clone.py
mpacer: Thank you!

mpacer:
I think the only outstanding issue is the question around `s3_key` vs `s3_object_key` and all of that if you think we should make the change, I'm happy to make it. 

It's definitely better to make it before this release than after :)

mpacer:
oh oops… i shouldn't have changed the api url in the template… that was me just being dumb…I'll push a new PR. I should have trusted the test :/

bookstore/clone.py
willingc: It's fine to leave as is. I agree re: passing around credentials.

I guess I am just trying to determine if there is any difference between the s3_key in the example GET request and the s3_object_key which is used internally by the method.

willingc:
No need to make a change. Just merge when it is green again. Nice job 👍

bookstore/clone.py
mpacer: There isn't… but we have the `s3_key` function in the `s3_paths` module, which is why we have the comment above the declaration. Is there any other way you would suggest conveying that?

```
        # s3_paths module has an s3_key function; s3_object_key avoids confusion
        s3_object_key = self.get_argument("s3_key")
```

mpacer:
Yay it's green! and thank you for the guidance — I'm excited to rebase the POST onto this, as I think that will massively improve test coverage.

Issue #122, opened 2019-06-07
Python API docs are broken/redundant


mpacer:
Right now the readthedocs doesn't show any autodoc documentation being built for the Python API: 
![image](https://user-images.githubusercontent.com/2482408/59134461-23965100-8930-11e9-8683-b6258c5f6714.png)

This is a fairly straightforward RST issue (number of underscore characters underneathe one of the rst header names in the store_client.rst).

But in addition, it looks like the methods for the handlers are being displayed twice, once from the class docstring's `Methods` section, once from each of the method's functional docstrings. The latter do not include the self parameter:
![image](https://user-images.githubusercontent.com/2482408/59134575-7ff97080-8930-11e9-8505-a408d3f63298.png)

@willingc What is the right way to go here to make the documentation make sense (ignoring the lack of the docs build which I'll fix with a small PR shortly).

willingc:
#123 (Thank you) fixed a problem. There is something up with the RTD build and the latest Sphinx changes and RTD.

I think the duplication is coming from the Class Docstring being followed closely by the method docstrings. We likely need to add some distinction either via css or headings.

willingc:
#124 fixed the build configuration. Working on polishing the docs in a separate PR.

willingc:
Hi @mpacer. I merged two PRs cleaning up the doc build and the formatting. I've added some additional headings on the autodoc pages. Please take a [look](https://bookstore.readthedocs.io/en/latest/reference/index.html). If you have any desired changes, either leave a note here, edit directly, or open a PR. Thanks!

willingc:
Going to close this @mpacer since I think this is resolved.

PR #124, opened 2019-06-08
Fix autodoc on rtd

willingc:
Addresses #122 

[Test build](https://test-bookstore.readthedocs.io/en/dociss/reference/archive.html#module-bookstore.archive)


willingc:
@mpacer merging so that the docs are displaying autodoc. We can finesse the formatting in a future PR.

PR #125, opened 2019-06-08
Use preferred form of traitlets integer

willingc:
Change `Int` to the preferred `Integer`. [Traitlets doc](https://traitlets.readthedocs.io/en/stable/trait_types.html#traitlets.Long)


mpacer:
Looks good to me!

PR #126, opened 2019-06-08
Reformat autodoc reference pages

willingc:
Addresses #122 

Adds headings and subheadings


willingc:
@mpacer I'm going to merge this as tests are green.

PR #128, opened 2019-06-08
Add test stubs for archive and publish

willingc:
There's still work to do with the async tests. I don't plan to add to this PR so please review as is.

### DIFF
--- a/bookstore/tests/test_publish.py
+++ b/bookstore/tests/test_publish.py
@@ -1,4 +1,5 @@
 import pytest
 
+
 def test_hello():
     assert 1000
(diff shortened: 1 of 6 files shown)


bookstore/tests/test_publish.py
```diff
@@ -0,0 +1,5 @@
+import pytest
+
+
+def test_hello():
+    assert 1000
```
MSeal: This seem erroneous?

willingc: Ah...this was to keep pytest from complaining about the empty file. So I threw in this test which should get removed when we have real publish tests ;-)

willingc: Let me remove and replace with something.

MSeal:
Thanks for adjusting the tests

PR #130, opened 2019-06-11
Add test VersionHandler; Update Version Handler's API

mpacer:
This refactors a bit of the data manipulation code in the handlers into individually testable utils.

This adds tests for the BookstoreVersionHandler.

There are a lot of issues that became obvious as I wrote these tests & helper utils. I will create separate issues for them, but I wanted to list them here. 

1. We should move BookstoreVersionHandler outside of the `handlers.py` module, into it's own module.
2. We could simplify the code without much work. But, the response is being used for validation by front-ends (e.g., nteract_on_jupyter), and so we will need to make sure that this works for them as well. 
Specifically, this would involve removing the `"bookstore": True` bit in the response from the BookstoreVersionHandler
3. Our tests are not all correctly awaiting the asynchronous handler methods, this means our synchronous code in those methods should be tested appropriately, but we'll need to add some more stuff to the other tests to get that working.

### DIFF
--- a/bookstore/handlers.py
+++ b/bookstore/handlers.py
@@ -41,13 +41,13 @@ class BookstoreVersionHandler(APIHandler):
         """Helper for building the version handler's response before serialization."""
         return {
             "release": self.settings['bookstore']["release"],
-            "validation": self.settings['bookstore']["validation"],
+            "features": self.settings['bookstore']["features"],
         }
 
 
 def build_settings_dict(validation):
     """Helper for building the settings info that will be assigned to the web_app."""
-    return {"release": version, "validation": validation}
+    return {"release": version, "features": validation}
 
 
 def load_jupyter_server_extension(nb_app):
--- a/bookstore/handlers.py
+++ b/bookstore/handlers.py
@@ -34,8 +34,8 @@ class BookstoreVersionHandler(APIHandler):
         }
 
 
-def build_settings_dict(bookstore_settings):
-    return {"version": version, "validation": validate_bookstore(bookstore_settings)}
+def build_settings_dict(validation):
+    return {"version": version, "validation": validation}
 
 
 def load_jupyter_server_extension(nb_app):
@@ -46,7 +46,7 @@ def load_jupyter_server_extension(nb_app):
 
     bookstore_settings = BookstoreSettings(parent=nb_app)
     validation = validate_bookstore(bookstore_settings)
-    web_app.settings['bookstore'] = {"version": version, "validation": validation}
+    web_app.settings['bookstore'] = build_settings_dict(validation)
     handlers = collect_handlers(nb_app.log, base_url, validation)
     web_app.add_handlers(host_pattern, handlers)
 
--- a/bookstore/handlers.py
+++ b/bookstore/handlers.py
@@ -40,15 +40,14 @@ class BookstoreVersionHandler(APIHandler):
     def build_response_dict(self):
         """Helper for building the version handler's response before serialization."""
         return {
-            "bookstore": True,  # TODO: Should we remove this; isn't it equivalent to the endpoint responding?
-            "version": self.settings['bookstore']["version"],
+            "release": self.settings['bookstore']["release"],
             "validation": self.settings['bookstore']["validation"],
         }
 
 
 def build_settings_dict(validation):
     """Helper for building the settings info that will be assigned to the web_app."""
-    return {"version": version, "validation": validation}
+    return {"release": version, "validation": validation}
 
 
 def load_jupyter_server_extension(nb_app):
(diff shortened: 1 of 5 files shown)


bookstore/handlers.py
```diff
@@ -20,24 +20,35 @@ class BookstoreVersionHandler(APIHandler):
...
+        """GET /api/bookstore/
+
+        Returns version info and validation info for various bookstore features.
+        """
+        self.finish(json.dumps(self.build_response_dict()))
+
+    def build_response_dict(self):
+        """Helper for building the version handler's response before serialization."""
+        return {
+            "bookstore": True,  # TODO: Should we remove this; isn't it equivalent to the endpoint responding?
```
MSeal: I don't know what this field even represents. What is it used for?

bookstore/tests/test_handlers.py
```diff
@@ -1,6 +1,87 @@
...
+            connection=connection,
+        )
+        return BookstoreVersionHandler(app, payload_request)
+
+    def test_get(self):
+        """This is a simple test of the get API at /api/bookstore
+
+        The most notable feature is the need to set _transforms on the handler.
+
+        The default value of handler()._transforms is `None`.
```
MSeal: Probably didn't need this much explanation

bookstore/tests/test_handlers.py
```diff
@@ -1,6 +1,87 @@
...
+        In normal usage, the application assigns this when it creates a handler delegate.
+
+        Because our mock application does not do this 
+        As a result this raises an error when self.finish() (and therefore self.flush()) is called.
+
+        At runtime on a live Jupyter server, application.transforms == [].
+        """
+        get_handler = self.get_handler('/api/bookstore/')
+        setattr(get_handler, '_transforms', [])
+        get_handler.get()
```
MSeal: We don't assert anything. There's a lot of code issues that would appear to pass this test when they are in fact not processing the get() call

bookstore/tests/test_handlers.py
```diff
@@ -1,6 +1,87 @@
...
 
-def test_handlers():
-    pass
+    def test_build_response(self):
+        empty_handler = self.get_handler('/api/bookstore/')
+        expected = {
+            'bookstore': True,
+            'validation': {'archive_valid': True, 'bookstore_valid': False, 'publish_valid': True},
+            'version': '2.3.0.dev',
```
MSeal: Will this test have to be updated each time the version is bumped? Might want to import the version for building this test object

bookstore/handlers.py
mpacer: I think it was originally a placeholder as a response indicating that indeed the endpoint was from bookstore. 

I want to tackle that in a different PR though. Check out & comment on #132 please!

bookstore/tests/test_handlers.py
mpacer: Given that we don't have other examples of this kind of a mocking tornado handler testing framework without a live application, this will matter for anyone else creating a new test. 

Perhaps we move the info into docs on testing.

But, I don't see any harm in being explicit about the reasoning given the seeming arbitrariness of this step.

bookstore/tests/test_handlers.py
mpacer: We're not doing this with all of the other handler tests. I agree it would be better, but I am not aiming at ideal right now.

bookstore/tests/test_handlers.py
mpacer: That's an excellent suggestion!

bookstore/handlers.py
```diff
@@ -20,24 +20,35 @@ class BookstoreVersionHandler(APIHandler):
...
+            "version": self.settings['bookstore']["version"],
```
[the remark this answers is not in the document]
mpacer: This has to do with the meaning of the handler, which I want to address in other PRs. This was only for the purpose of adding testing to what we currently have.

We could expand the scope of the #132 conversation to be about reconsidering this API's content altogether.

bookstore/tests/test_handlers.py
mpacer: What are those code issues?

mpacer:
Hi @MSeal @willingc I made the changes to the API that you suggested as part of this PR now. 

this has been rebased on master & is ready for the next round of reviews.

bookstore/handlers.py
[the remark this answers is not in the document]
mpacer: I made this change in bfeecf1.

bookstore/handlers.py
mpacer: I made this change in bfeecf1

bookstore/tests/test_handlers.py
MSeal: What if the get returns None? Or get_handler returns the post handler? This test as is doesn't confirm much.

mpacer: I think the get _should_ return `None` given the way that tornado handlers are defined (it treats the finish as a side-effect that I'm not sure how to hook into). 

What it confirms is that the code runs without raising an error, meaning we've met the API for the `get` method, and that such a handler can be created without issue.

The problem is there is no way that I can seem to get access to whatever is sent over the wire based on get.

mpacer:
@willingc I'm sorry I missed your comment about changing the name of the top level property of the API response from "validation" to "features" — I just caught it and made the change

willingc:
I don't see any outstanding issues in this PR so I am going to merge. Thanks @mpacer 🌻

mpacer:
Thanks!

PR #134, opened 2019-06-15
Refactor & Test Validation and Handler Adding

mpacer:
The biggest win in this is that this tests which endpoints & handlers are actually enabled in the handlers module. #130 does not address this, and should probably be rebased on top of this if this is merged.

- adds a way to disable cloning via `BookstoreSettings.cloning_enabled` traitlet
- adds `cloning_enabled` to validation code
- adds a new `build_handlers` function so we can test it directly (allowing less mocking)
- changes the semantics of `publish_enabled` and `archive_enabled`
- updates our `pytest` requirement to include `pytest>=3.6` (as required by `pytest-asyncio` as reported in [this issue](pytest-dev/pytest-asyncio#115))

### DIFF
--- a/bookstore/handlers.py
+++ b/bookstore/handlers.py
@@ -56,7 +56,7 @@ def load_jupyter_server_extension(nb_app):
 def build_handlers(log, base_url, validation):
     """Utility for determinine which handlers should be added to the webapp.
 
-    This returns both the paths that are to be added as well as 
+    This returns each endpoint path and its associated handler.
 
     Parameters
     ----------
@@ -77,7 +77,7 @@ def build_handlers(log, base_url, validation):
     base_bookstore_api_pattern = url_path_join(base_url, '/api/bookstore')
 
     handlers = []
-    # Always enable the version handler
+    # Always enable the version handler for the API
     handlers.append((base_bookstore_api_pattern, BookstoreVersionHandler))
 
     if validation['publish_valid']:
@@ -90,7 +90,7 @@ def build_handlers(log, base_url, validation):
         )
     else:
         log.info(
-            "[bookstore] Not enabling bookstore publishing, s3_bucket or endpoint not configured"
+            "[bookstore] Publishing disabled. s3_bucket or endpoint are not configured."
         )
 
     if validation['cloning_valid']:
@@ -102,6 +102,6 @@ def build_handlers(log, base_url, validation):
             (url_path_join(base_bookstore_pattern, r"/clone(?:/?)*"), BookstoreCloneHandler)
         )
     else:
-        log.info(f"[bookstore] Not enabling bookstore cloning, version: {version}")
+        log.info(f"[bookstore] bookstore cloning disabled, version: {version}")
 
     return handlers
--- a/bookstore/handlers.py
+++ b/bookstore/handlers.py
@@ -49,17 +49,10 @@ def load_jupyter_server_extension(nb_app):
     base_bookstore_api_pattern = url_path_join(web_app.settings['base_url'], '/api/bookstore')
     web_app.add_handlers(host_pattern, [(base_bookstore_api_pattern, BookstoreVersionHandler)])
     bookstore_settings = BookstoreSettings(parent=nb_app)
-    web_app.settings['bookstore'] = {
-        "version": version,
-        "validation": validate_bookstore(bookstore_settings),
-    }
+    validation = validate_bookstore(bookstore_settings)
+    web_app.settings['bookstore'] = {"version": version, "validation": validation}
 
-    check_published = [
-        web_app.settings['bookstore']['validation'].get("bookstore_valid"),
-        web_app.settings['bookstore']['validation'].get("publish_valid"),
-    ]
-
-    if not all(check_published):
+    if not validation['publish_valid']:
         nb_app.log.info("[bookstore] Not enabling bookstore publishing, endpoint not configured")
     else:
         nb_app.log.info(f"[bookstore] Enabling bookstore publishing, version: {version}")
@@ -73,10 +66,14 @@ def load_jupyter_server_extension(nb_app):
             ],
         )
 
-    web_app.add_handlers(
-        host_pattern,
-        [
-            (url_path_join(base_bookstore_api_pattern, r"/clone(?:/?)*"), BookstoreCloneAPIHandler),
-            (url_path_join(base_bookstore_pattern, r"/clone(?:/?)*"), BookstoreCloneHandler),
-        ],
-    )
+    if validation['cloning_valid']:
+        web_app.add_handlers(
+            host_pattern,
+            [
+                (
+                    url_path_join(base_bookstore_api_pattern, r"/clone(?:/?)*"),
+                    BookstoreCloneAPIHandler,
+                ),
+                (url_path_join(base_bookstore_pattern, r"/clone(?:/?)*"), BookstoreCloneHandler),
+            ],
+        )
--- a/bookstore/handlers.py
+++ b/bookstore/handlers.py
@@ -49,14 +49,15 @@ def load_jupyter_server_extension(nb_app):
     bookstore_settings = BookstoreSettings(parent=nb_app)
     validation = validate_bookstore(bookstore_settings)
     web_app.settings['bookstore'] = {"version": version, "validation": validation}
-    handlers = build_handlers(nb_app.log, base_url, validation)
+    handlers = collect_handlers(nb_app.log, base_url, validation)
     web_app.add_handlers(host_pattern, handlers)
 
 
-def build_handlers(log, base_url, validation):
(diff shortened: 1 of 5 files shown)


mpacer:
I'd like to move the code that defines which handlers are to be added into the respective modules as well (clone & publish). But I figured this is already touching enough files that it might be a lot to review.

bookstore/handlers.py
```diff
@@ -44,39 +44,64 @@ def load_jupyter_server_extension(nb_app):
...
-
-    if not all(check_published):
-        nb_app.log.info("[bookstore] Not enabling bookstore publishing, endpoint not configured")
+    validation = validate_bookstore(bookstore_settings)
+    web_app.settings['bookstore'] = {"version": version, "validation": validation}
+    handlers = build_handlers(nb_app.log, base_url, validation)
+    web_app.add_handlers(host_pattern, handlers)
+
+
+def build_handlers(log, base_url, validation):
```
willingc: Perhaps rename to `webapp_handlers` or `base_handlers`.

bookstore/handlers.py
```diff
@@ -44,39 +44,64 @@ def load_jupyter_server_extension(nb_app):
...
+    """Utility for determinine which handlers should be added to the webapp.
```
willingc:
```suggestion
    """Utility for determine which handlers should be added to the webapp.
```

bookstore/handlers.py
```diff
@@ -44,39 +44,64 @@ def load_jupyter_server_extension(nb_app):
...
+    base_url: str
+      The base_url to which we append routes.
+    validation: dict
+       Validation dictionary for determining which endpoints to enable.
+
+    Returns
+    --------
+    
+    List[Tuple[str, tornado.web.RequestHandler]]
+        Template parameters in a dictionary
```
willingc: list or dictionary?

bookstore/handlers.py
```diff
@@ -44,39 +44,64 @@ def load_jupyter_server_extension(nb_app):
...
+    validation = validate_bookstore(bookstore_settings)
+    web_app.settings['bookstore'] = {"version": version, "validation": validation}
+    handlers = build_handlers(nb_app.log, base_url, validation)
+    web_app.add_handlers(host_pattern, handlers)
+
+
+def build_handlers(log, base_url, validation):
+    """Utility for determinine which handlers should be added to the webapp.
+
+    This returns both the paths that are to be added as well as
```
willingc:
```suggestion
    This returns each endpoint path and its associated handler.
```

bookstore/handlers.py
```diff
@@ -44,39 +44,64 @@ def load_jupyter_server_extension(nb_app):
...
+    --------
+    
+    List[Tuple[str, tornado.web.RequestHandler]]
+        Template parameters in a dictionary
+    """
+    base_bookstore_pattern = url_path_join(base_url, '/bookstore')
+    base_bookstore_api_pattern = url_path_join(base_url, '/api/bookstore')
+
+    handlers = []
+    # Always enable the version handler
```
willingc:
```suggestion
    # Always enable the version handler for the API
```

bookstore/handlers.py
```diff
@@ -44,39 +44,64 @@ def load_jupyter_server_extension(nb_app):
...
-        web_app.add_handlers(
-            host_pattern,
-            [
-                (
-                    url_path_join(base_bookstore_api_pattern, r"/publish%s" % path_regex),
-                    BookstorePublishHandler,
-                )
-            ],
+        log.info(
+            "[bookstore] Not enabling bookstore publishing, s3_bucket or endpoint not configured"
```
willingc:
```suggestion
            "[bookstore] Publishing disabled. s3_bucket or endpoint are not configured."
```

bookstore/handlers.py
```diff
@@ -44,39 +44,64 @@ def load_jupyter_server_extension(nb_app):
...
+    if validation['cloning_valid']:
+        log.info(f"[bookstore] Enabling bookstore cloning, version: {version}")
+        handlers.append(
+            (url_path_join(base_bookstore_api_pattern, r"/clone(?:/?)*"), BookstoreCloneAPIHandler)
+        ),
+        handlers.append(
+            (url_path_join(base_bookstore_pattern, r"/clone(?:/?)*"), BookstoreCloneHandler)
+        )
+    else:
+        log.info(f"[bookstore] Not enabling bookstore cloning, version: {version}")
```
willingc:
```suggestion
        log.info(f"[bookstore] bookstore cloning disabled, version: {version}")
```

bookstore/tests/test_bookstore_config.py
```diff
@@ -6,31 +6,71 @@
...
+        "bookstore_valid": False,
+        "publish_valid": False,
+        "archive_valid": False,
+        "cloning_valid": True,
+    }
     settings = BookstoreSettings()
-    assert not validate_bookstore(settings)['bookstore_valid']
-    assert validate_bookstore(settings)['publish_valid']
-    assert validate_bookstore(settings)['archive_valid']
+    assert validate_bookstore(settings) == expected
```
willingc: Nice refactor :D

bookstore/tests/test_handlers.py
```diff
@@ -1,6 +1,76 @@
 """Tests for handlers"""
 import pytest
+import logging
+
```
willingc:
```suggestion
```

bookstore/tests/test_handlers.py
```diff
@@ -1,6 +1,76 @@
...
+from unittest.mock import Mock
+
+from bookstore.handlers import build_handlers, BookstoreVersionHandler
+from bookstore.bookstore_config import BookstoreSettings, validate_bookstore
+from bookstore.clone import BookstoreCloneHandler, BookstoreCloneAPIHandler
+from bookstore.publish import BookstorePublishHandler
+
```
willingc:
```suggestion
```

bookstore/tests/test_handlers.py
```diff
@@ -1,6 +1,76 @@
...
     pass
+
+
+def test_build_handlers_all():
+    web_app = Application()
+    mock_settings = {"BookstoreSettings": {"s3_bucket": "mock_bucket"}}
+    bookstore_settings = BookstoreSettings(config=Config(mock_settings))
+    validation = validate_bookstore(bookstore_settings)
+    handlers = build_handlers(log, '/', validation)
+    expected = [
```
willingc: Probably would be better to put the expected result before the test steps. Expected result. Run the test. Assert test results match expected.

bookstore/handlers.py
mpacer: It returns a list of tuples of the form (<endpoint_pattern>, <handler_class>).

bookstore/tests/test_handlers.py
mpacer: Super reasonable…and usually the pattern I prefer — I don't know what I was thinking.

bookstore/handlers.py
mpacer: I think it needs to be "for determining" or "to determine", but I'll just accept this as is and fix that in a follow-up commit.

bookstore/handlers.py
mpacer: Would `collect_handlers` work? Either would make sense to me as a property name on a class, but I normally try to stick with verb-y names for functions & methods & noun-y names for properties/attributes/classes. 

The tricky thing is that it's not itself adding them to the web_app & it feels odd to call it base_handlers when it's not a constant set of handlers (i.e., it'd feel more appropriate if we were only getting the "version" handler here).

bookstore/handlers.py
willingc: Any reason to keep dictionary in the comment?

bookstore/handlers.py
willingc: I like collect_handlers.

bookstore/tests/test_handlers.py
willingc: It's Friday :wink:

willingc:
Feel free to merge @mpacer.

bookstore/handlers.py
mpacer: Nope — just a carry-over from when I copied the syntax for the returns method from another location. Totally blanked on that one — sorry!

bookstore/tests/test_bookstore_config.py
mpacer: Grazie!

mpacer:
Grazie! @willingc, I made the followup changes after merging your suggestions. Is there anything more that this needs before we can merge it?

mpacer:
Definitely didn't forget to change the method name everywhere causing the handlers to never be added to the web_app & all integration tests to fail… nope definitely wouldn't have done that. ;)

mpacer:
Alright, merging!

PR #137, opened 2019-06-27
Publish test

mpacer:
This PR grew a bit, however the core of it is the creation of new publish handler tests.

It also refactors some of the code so it is more encapsulated and testable.

Because I realized as I was going through the tests that we weren't doing a great job of catching invalid request bodies, It also adds better validation around those.

That then led to a realization that our API docs were not appropriately specific about the request body.
Specifically, it said it needed to be a Contents API compatible message, but that wasn't the case. Furthermore, that behaviour doesn't match the APIs expected in the current notebook's contents handlers.

### DIFF
--- a/bookstore/publish.py
+++ b/bookstore/publish.py
@@ -2,7 +2,7 @@ import json
 
 import aiobotocore
 
-
+from botocore.exceptions import ClientError
 from nbformat import ValidationError, validate as validate_nb
 from notebook.base.handlers import APIHandler, path_regex
 from notebook.services.contents.handlers import validate_model
@@ -90,11 +90,15 @@ class BookstorePublishAPIHandler(APIHandler):
             region_name=self.bookstore_settings.s3_region_name,
         ) as client:
             self.log.info("Processing published write of %s", path)
-            obj = await client.put_object(
-                Bucket=self.bookstore_settings.s3_bucket,
-                Key=s3_object_key,
-                Body=json.dumps(content),
-            )
+            try:
+                obj = await client.put_object(
+                    Bucket=self.bookstore_settings.s3_bucket,
+                    Key=s3_object_key,
+                    Body=json.dumps(content),
+                )
+            except ClientError as e:
+                status_code = e.response['ResponseMetadata'].get('HTTPStatusCode')
+                raise web.HTTPError(status_code, e.args[0])
             self.log.info("Done with published write of %s", path)
 
         resp_content = self.prepare_response(path, obj)
--- a/bookstore/publish.py
+++ b/bookstore/publish.py
@@ -11,7 +11,7 @@ from .s3_paths import s3_display_path
 from .utils import url_path_join
 
 
-class BookstorePublishHandler(APIHandler):
+class BookstorePublishAPIHandler(APIHandler):
     """Publish a notebook to the publish path"""
 
     def initialize(self):
--- a/bookstore/publish.py
+++ b/bookstore/publish.py
@@ -45,10 +45,14 @@ class BookstorePublishAPIHandler(APIHandler):
         self.finish(resp)
 
     def validate_model(self, model):
-        """This is a helper that validates that the model meets the notebook Contents API.
+        """This is a helper that validates that the model meets our expected structure.
 
-        It uses the validate_model method provided in the contents API handler.
+        Raises
+        ------
+        tornado.web.HTTPError
+            Your model does not validate correctly
         """
+        # TODO: This is far more lenient than our API docs would suggest. We should reconcile them.
         if model['type'] != 'notebook':
             raise web.HTTPError(400, "bookstore only publishes notebooks")
 
--- a/bookstore/publish.py
+++ b/bookstore/publish.py
@@ -43,7 +43,7 @@ class BookstorePublishAPIHandler(APIHandler):
         self.finish(resp)
 
     def validate_model(self, model):
-        """This is a helper that validates that the model meets our expected structure.
+        """Checks that the model given to the API handler meets bookstore's expected structure for a notebook.
 
         Pattern for surfacing nbformat validation errors originally written in
         https://github.com/jupyter/notebook/blob/a44a367c219b60a19bee003877d32c3ff1ce2412/notebook/services/contents/manager.py#L353-L355
--- a/bookstore/publish.py
+++ b/bookstore/publish.py
@@ -1,6 +1,9 @@
 import json
 
 import aiobotocore
+
+
+from nbformat import ValidationError, validate as validate_nb
 from notebook.base.handlers import APIHandler, path_regex
 from notebook.services.contents.handlers import validate_model
 from tornado import web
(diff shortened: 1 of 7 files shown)


bookstore/publish.py
```diff
@@ -33,20 +37,42 @@ def initialize(self):
             raise web.HTTPError(400, "Must provide a path for publishing")
 
         model = self.get_json_body()
-        if model:
-            await self._publish(model, path.lstrip('/'))
-        else:
-            raise web.HTTPError(400, "Cannot publish an empty model")
+        self.validate_model(model)
+        resp = await self._publish(model['content'], path.lstrip('/'))
+        self.set_status(201)
```
willingc: Is there a particular reason that we are setting the status as 201 instead of taking from the response?

bookstore/publish.py
```diff
@@ -33,20 +37,42 @@ def initialize(self):
...
+        self.finish(resp)
 
-    async def _publish(self, model, path):
-        """Publish notebook model to the path"""
-        if model['type'] != 'notebook':
-            raise web.HTTPError(400, "bookstore only publishes notebooks")
-        content = model['content']
+    def validate_model(self, model):
+        """This is a helper that validates that the model meets our expected structure.
```
willingc:
```suggestion
        """Checks that the model given to the API handler meets bookstore's expected structure for a notebook.
```

bookstore/publish.py
```diff
@@ -64,19 +90,30 @@ def initialize(self):
...
+                    Body=json.dumps(content),
+                )
+            except ClientError as e:
+                status_code = e.response['ResponseMetadata'].get('HTTPStatusCode')
+                raise web.HTTPError(status_code, e.args[0])
             self.log.info("Done with published write of %s", path)
 
-        self.set_status(201)
+        resp_content = self.prepare_response(path, obj)
+        resp_str = json.dumps(resp_content)
```
willingc: Perhaps:

`return json.dumps(resp_content)`

bookstore/tests/test_publish.py
```diff
@@ -1,9 +1,109 @@
...
+            uri=uri,
+            headers={"Host": "localhost:8888"},
+            body=body,
+            connection=connection,
+        )
+        return BookstorePublishAPIHandler(app, payload_request)
+
+    @gen_test
+    async def test_put_no_path(self):
+        empty_handler = self.put_handler('/bookstore/publish/')
```
willingc: @mpacer I'm not quite clear what you mean by  `empty_handler`.

bookstore/tests/test_publish.py
```diff
@@ -1,9 +1,109 @@
...
+from traitlets.config import Config
 
 
 def test_create_publish_handler_no_params():
     with pytest.raises(TypeError):
-        assert BookstorePublishHandler()
+        assert BookstorePublishAPIHandler()
+
+
+class TestPublishAPIHandler(AsyncTestCase):
```
willingc: Perhaps rename class to: `MockPublishAPIHandler(AsyncTestCase)`

willingc: I'm now seeing that the tests are done within the class. @mpacer @MSeal, is the typical pattern to do it this way with S3? I would have expected creating a MockPublishAPIHandler and then running the tests outside of the Mock handler.

bookstore/publish.py
```diff
@@ -1,7 +1,11 @@
 import json
 
 import aiobotocore
+
+from botocore.exceptions import ClientError
+from nbformat import ValidationError, validate as validate_nb
```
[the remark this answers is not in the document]
mpacer: Sounds great!

bookstore/publish.py
mpacer: Do you mean setting the status to match the response from the S3 client? We can do that… we hadn't been doing that for the error handling (because from our server's perspective we didn't have the same kind of error)… but I suppose the successful case can be passed through.

But for now I'll need to dive into the code to see what the actual response is.

bookstore/publish.py
mpacer: that's what I'd originally had but was probably trying to overoptimize for new contributor legibility.

bookstore/tests/test_publish.py
mpacer: This is how we're avoiding the whole s3 mocking while still unit testing our handlers. This comes from the model given by tornado. In these cases we're needing to build the handler slightly differently each time because of how it is instantiated on the Tornado side.

bookstore/tests/test_publish.py
mpacer: i can change it to `no_path_handler`

bookstore/tests/test_publish.py
mpacer: figuring out this model & how to work with it honestly has been the hardest conceptual part of the entire unit-testing project, as we hadn't found anyone else that was doing true unit tests on their handlers (i.e., without making a mock server, without handling routing through a real application, without sending any actual http requests, &c.).

bookstore/publish.py
willingc: Fair enough. I typically lean toward less lines is less code to maintain after understanding. In this particular case, either is fine. If the lines were several away from each other, I would lean toward being explicit to avoid confusion. Since they are close together, I tend to opt to leave out the placeholder variable.

bookstore/tests/test_publish.py
willingc: Makes sense. We create a Mock server on JupyterHub even though we are using Tornado. If you and Matt are in agreement on the approach, I'm happy with it. It was more curiosity on my part.

willingc:
Thanks for the changes @mpacer and the explanations. I'm good with the PR as it stands. I will leave for Matthew to review and merge. :sunny:

bookstore/tests/test_publish.py
mpacer: i'd tried to follow through the JupyterHub code base & got a lot of really great pointers on how to approach this, but ended up not being able to use too many of the patterns since it did run a a server to handle the requests.

I'm definitely open to taking that approach in the future, if you'd like — it would probably simply a lot of the weirder workarounds we've been doing at the sacrifice of making our unit tests a little more integration-y.

If we wanted to do that, I would really appreciate your guidance around the JupyterHub testing base, because while I was able to learn a lot… its really complicated & was tricky to see how we'd adapt it to our use cases.

bookstore/publish.py
```diff
@@ -33,20 +38,41 @@ def initialize(self):
...
+            Your model does not validate correctly
+        """
+        if not model:
+            raise web.HTTPError(400, "Bookstore cannot publish an empty model")
+        if model.get('type', "") != 'notebook':
+            raise web.HTTPError(400, "Bookstore only publishes notebooks")
+
+        content = model.get('content', {})
+        if content == {}:
+            raise web.HTTPError(400, "Bookstore cannot publish empty contents")
```
MSeal: Minor, but should be a 422 (https://www.restapitutorial.com/httpstatuscodes.html)

bookstore/publish.py
```diff
@@ -33,20 +38,41 @@ def initialize(self):
...
+
+        Raises
+        ------
+        tornado.web.HTTPError
+            Your model does not validate correctly
+        """
+        if not model:
+            raise web.HTTPError(400, "Bookstore cannot publish an empty model")
+        if model.get('type', "") != 'notebook':
+            raise web.HTTPError(400, "Bookstore only publishes notebooks")
```
MSeal: 415 status code

bookstore/publish.py
```diff
@@ -33,20 +38,41 @@ def initialize(self):
...
+
+        content = model.get('content', {})
+        if content == {}:
+            raise web.HTTPError(400, "Bookstore cannot publish empty contents")
+        try:
+            validate_nb(content)
+        except ValidationError as e:
+            raise web.HTTPError(
+                400,
```
MSeal: 422 status code as well is generally more accepted for validation errors (see https://stackoverflow.com/questions/3290182/rest-http-status-codes-for-failed-validation-or-invalid-duplicate for some discussions on the topic). 400 is probably second best for this so if you want to just leave these as 400 it'd be ok

bookstore/tests/test_publish.py
MSeal: I'm good with this approach. There's still a few patterns like this where it's awkward / difficult to get in a pure pytest pattern. We could probably reverse engineer a complicated fixture / context wrapper pattern to emulate all the pieces, but given the work required to get the tests running at all in this repo I don't think it'd be worth the effort to go through that additional pain.

bookstore/publish.py
mpacer: My understanding was that that was for literal media type conflicts (media type in the sense of "mimetypes").

In the contents API that would be on the "format" field, not the "type" field. But since we're not following the full contents API here, I'm ok making this change

bookstore/publish.py
mpacer: I can make the change, but I would have thought that we were expecting 'content' on the request — I suppose the tricky thing is that this check both would trigger if content is not present on the request (400) and if it is present but is an empty dict (422)…

Is it worth making that distinction?

mpacer:
Merging, thank you @MSeal and @willingc for your reviews!!

PR #138, opened 2019-06-28
Archive test improvements

mpacer:
From my perspective, the most important thing this PR does is to add a test to ensure that we test the code path where we skip archiving based on a lock having been already acquired.

### DIFF
--- a/bookstore/archive.py
+++ b/bookstore/archive.py
@@ -145,6 +145,7 @@ class BookstoreContentsArchiver(FileContentsManager):
             The storage location
         """
         if model["type"] != "notebook":
+            self.log.info("Bookstore only archives notebooks.")
             return
 
         content = json.dumps(model["content"])
--- a/bookstore/archive.py
+++ b/bookstore/archive.py
@@ -139,8 +139,8 @@ class BookstoreContentsArchiver(FileContentsManager):
         Parameters
         ----------
 
-        model : str
-            The type of file
+        model : dict
+            The type of file and its contents
         path : str
             The storage location
         """
--- a/bookstore/archive.py
+++ b/bookstore/archive.py
@@ -145,7 +145,10 @@ class BookstoreContentsArchiver(FileContentsManager):
             The storage location
         """
         if model["type"] != "notebook":
-            self.log.info("Bookstore only archives notebooks.")
+            self.log.debug(
+                "Bookstore only archives notebooks, "
+                f"request does not state that {path} is a notebook."
+            )
             return
 
         content = json.dumps(model["content"])
--- a/bookstore/tests/test_archive.py
+++ b/bookstore/tests/test_archive.py
@@ -2,6 +2,7 @@
 import asyncio
 import pytest
 import json
+import logging
 
 from bookstore.archive import ArchiveRecord, BookstoreContentsArchiver
 from nbformat.v4 import new_notebook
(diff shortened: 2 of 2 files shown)


bookstore/archive.py
```diff
@@ -139,12 +139,13 @@ def run_pre_save_hook(self, model, path, **kwargs):
         Parameters
         ----------
 
-        model : str
-            The type of file
+        model : dict
+            The type of file and its contents
```
rgbkrk: 👍

bookstore/archive.py
```diff
@@ -139,12 +139,13 @@ def run_pre_save_hook(self, model, path, **kwargs):
...
         path : str
             The storage location
         """
         if model["type"] != "notebook":
+            self.log.info("Bookstore only archives notebooks.")
```
rgbkrk: Thinking about what this will look like in the logs, if someone is editing any file, it's going to show up as:

```
[I 16:46:40.790 NotebookApp] Bookstore only archives notebooks.
[I 16:48:40.790 NotebookApp] Bookstore only archives notebooks.
[I 16:50:40.790 NotebookApp] Bookstore only archives notebooks.
```

so I'd recommend setting this to `self.log.debug` and/or adding the filename to the message so this isn't useless amidst larger user logs (for operators to sift through).

bookstore/tests/test_archive.py
```diff
@@ -19,7 +21,34 @@ def test_create_contentsarchiver_invalid_args_count():
...
+    record = ArchiveRecord('my_notebook_path.ipynb', json.dumps(new_notebook()), 100.2)
     assert record
 
     await archiver.archive(record)
+
+
+@pytest.mark.asyncio
+async def test_archive_abort_with_lock():
+    archiver = BookstoreContentsArchiver()
+    record = ArchiveRecord('my_notebook_path.ipynb', json.dumps(new_notebook()), 100.2)
```
rgbkrk: Some extra comments I'd love to see added to this test:

```
Acquire a lock in advance so that when the archiver attempts to archive, it will abort.
```

Not sure how we can test to make sure that it really did abort here -- mocking the log? Returning a boolean? Verifying that `put_object` wasn't called by mocking s3?

bookstore/archive.py
mpacer: That's a really good point. I'll make it debug.

bookstore/tests/test_archive.py
mpacer: I just learned about pytests logging capture fixture (https://docs.pytest.org/en/latest/logging.html#caplog-fixture)… it makes the logs really easy to capture.

bookstore/archive.py
mpacer: I also added a line for specifying the file in question

mpacer:
@rgbkrk do you have any remaining concerns?

bookstore/tests/test_archive.py
```diff
@@ -19,7 +22,38 @@ def test_create_contentsarchiver_invalid_args_count():
...
+
+    archiver = BookstoreContentsArchiver()
+    record = ArchiveRecord('my_notebook_path.ipynb', json.dumps(new_notebook()), 100.2)
+
+    lock = asyncio.Lock()
+    archiver.path_locks['my_notebook_path.ipynb'] = lock
+    async with lock:
+        with caplog.at_level(logging.INFO):
+            await archiver.archive(record)
+    assert 'Skipping archive of my_notebook_path.ipynb' in caplog.text
```
rgbkrk: ❤️

mpacer:
> These tests are sweet.

I know! The caplog fixture is totally changing how I think about tests.

PR #141, opened 2019-06-29
Publish refactor

mpacer:
This takes some of the lessons from #139 and applies them to the publish module. 

Most notably it takes as much logic as possible out of the `_publish` method, making it only about speaking to s3 and nothing else.

Also, this moves a lot of the s3 string creation into the base `put` method & simplifies some of the logic that was happening in `prepare_response`.

This needed to build on #140 (since it changed the docstring), so if this is merged, that can be closed.

### DIFF
--- a/bookstore/publish.py
+++ b/bookstore/publish.py
@@ -37,15 +37,25 @@ class BookstorePublishAPIHandler(APIHandler):
         path: str
             Path describing where contents should be published to, postfixed to the published_prefix .
         """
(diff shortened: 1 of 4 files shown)


mpacer:
for some reason the code coverage reports got messed up when I rebased on master… I'm not entirely sure why they're saying it reduces coverage… it doesn't change coverage at all

mpacer:
Grazie!

PR #142, opened 2019-07-02
Fixes client endpoints and adds docstrings

mpacer:
As I was writing up the changelog, I realised we should be highlighting that in comparison to the rest of the library, these are experimental features that are not as well tested. 

For instance, I realized that the current endpoints for bookstore did not match those that we actually provide as part of our API 🤦‍♀.

The changes to the endpoints are the only non-doc changes. If needed to get this ready to be merged in time for 2.3.0, I can break those out into a separate PR before we release: https://github.com/nteract/bookstore/pull/142/commits/80a9559f5a0d50175919f0dfda74dcdfa116d5bb

So I fixed that, and updated all of their docstrings to mention their experimental nature. I didn't do so for the records because we're actually testing those. 

While I was at it I also improved the docstrings on the various class methods & modules.

### DIFF
--- a/bookstore/client/store_client.py
+++ b/bookstore/client/store_client.py
@@ -1,11 +1,23 @@
-"""Client to interact with the notebook store"""
+"""Client to interact with a notebook's bookstore functionality."""
 import requests
 
 from .nb_client import CurrentNotebookClient
 
 
 class BookstoreClient(CurrentNotebookClient):
-    """Represents a bookstore client that corresponds to the active nb client"""
+    """EXPERIMENTAL SUPPORT: A client that allows access to a Bookstore from within a notebook.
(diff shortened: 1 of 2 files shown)


bookstore/client/store_client.py
```diff
@@ -14,11 +26,20 @@ def __init__(self, s3_bucket=None):
 
     @property
     def publish_endpoint(self):
-        api_endpoint = "/api/bookstore/published/"
+        """Helper to refer to construct the publish endpoint for this notebook server."""
+        api_endpoint = "/api/bookstore/publish/"
```
willingc: Good catch!

mpacer:
Grazie!

Issue #145, opened 2019-07-10
Clone hanging when reading asyncio response object body from S3


mpacer:
Today, when I was using bookstore to clone from S3, I found that responses were just hanging indefinitely.

It seems to be occurring when we attempt to read the s3 response object's body (with `await obj['Body'].read()`).

Importantly, it doesn't hang when using minio's mock s3 server.

Does anyone have any insight as to what could be causing this?

@kylek @willingc @mseal

mpacer:
I figured out that this seems to be related to trying to read the Body outside of the client context manager. 

What's odd is that it still works with minio, and so I'm wondering if there is a hidden race condition.

In any case, I opened https://github.com/aio-libs/aiobotocore/issues/710 so we could track the underlying issue there.

In the meantime I'll work on a refactor so we can get 2.3.1 out the door.

PR #146, opened 2019-07-11
Move async content reading inside s3 client context manager

mpacer:
This addresses #145 for now. However it does not clarify why the issue arose in the first place.

It is likely that this results in a slightly lower coverage report. 

But, it has some advantages:
1. makes the tests slightly cleaner (since we no longer need to make a class to handle the asynchronous read) 
2. simplifies the API for build_content_model as we now can pass in a string rather than a dict with unused additional fields


mpacer:
@willingc are you comfortable with us merging this? There's not going to be an easy way to avoid dropping code coverage while fixing this bug :/.

MSeal:
-0.29% for refactoring isn't a problem worth blocking a merge imho

PR #156, opened 2019-08-01
Update clone handler template

mpacer:
As part of the fs cloning handler, it requires changing how the clone.html template works. This makes it easier to construct a template that works for both handlers.

Main two changes are to move common behaviour to the backend, specifically:

1. creating the post model
2. creating the description of the source file


mpacer:
@chyzzqo2 does this look ok to you?

PR #157, opened 2019-08-01
Validate relpath util

mpacer:
This creates the utility that will be shared between the fs cloning handler & the fs cloning api handler:

- checks the behaviour of the basedir 
- ensures relpath exists 
- ensures relpath does escape out of the basedir

### DIFF
--- a/bookstore/clone.py
+++ b/bookstore/clone.py
@@ -308,14 +308,17 @@ def validate_relpath(relpath, settings, log):
 
     Parameters
     ----------
-    model : dict
-        Jupyter Contents API model
-    obj : dict
+    relpath : string
+        Relative path to a notebook to be cloned.
+    settings : BookstoreSettings
+        Bookstore configuration.
+    log : logging.Logger
         Log (usually from the NotebookApp) for logging endpoint changes.
-    s3_bucket : str
-        The S3 bucket we are cloning from
-    s3_object_key: str
-        The S3 key we are cloning
+
+    Returns
+    --------
+    Path
+        Absolute path to file to be cloned.
     """
     if relpath == '':
         log.info("Request received with empty relpath.")
--- a/bookstore/clone.py
+++ b/bookstore/clone.py
@@ -303,19 +303,36 @@ class BookstoreCloneAPIHandler(APIHandler):
         return model
 
 
-def validate_relpath(relpath, settings):
+def validate_relpath(relpath, settings, log):
     """Validates that a relative path appropriately resolves given bookstore settings.
+
+    Parameters
+    ----------
+    model : dict
+        Jupyter Contents API model
+    obj : dict
+        Log (usually from the NotebookApp) for logging endpoint changes.
+    s3_bucket : str
+        The S3 bucket we are cloning from
+    s3_object_key: str
+        The S3 key we are cloning
     """
     if relpath == '':
-        raise web.HTTPError(400, "Requires a path to clone")
+        log.info("Request received with empty relpath.")
+        raise web.HTTPError(400, "Request malformed, must provide a non-empty relative path.")
 
     fs_basedir = Path(settings.fs_cloning_basedir)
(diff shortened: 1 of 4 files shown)


bookstore/clone.py
```diff
@@ -299,3 +300,21 @@ def build_post_response_model(self, model, obj, s3_bucket, s3_object_key):
         if 'VersionId' in obj:
             model["versionID"] = obj['VersionId']
         return model
+
+
+def validate_relpath(relpath, settings):
+    """Validates that a relative path appropriately resolves given bookstore settings.
+    """
+    if relpath == '':
```
chyzzqo2: The first two errors are server misconfigs. I would use a 501 and generically say. Server not configured for this. Details of why we think this is should be in server logs not response text.

bookstore/clone.py
```diff
@@ -299,3 +300,21 @@ def build_post_response_model(self, model, obj, s3_bucket, s3_object_key):
...
+        raise web.HTTPError(400, "Requires a path to clone")
+
+    fs_basedir = Path(settings.fs_cloning_basedir)
+    if not fs_basedir.is_absolute():
+        raise web.HTTPError(400, "File system cloning base directory must be an absolute path")
```
chyzzqo2: Same as a above

bookstore/clone.py
```diff
@@ -299,3 +300,21 @@ def build_post_response_model(self, model, obj, s3_bucket, s3_object_key):
...
+
+    fs_clonepath = Path(os.path.realpath(os.path.join(fs_basedir, relpath)))
+
+    if fs_basedir not in fs_clonepath.parents:
+        raise web.HTTPError(400, "Cannot clone from a path outside of the base directory")
```
chyzzqo2: Suggest making return the same http response as if the if the file did not exist, with only log messages that clarify why we rejected it

bookstore/clone.py
mpacer: Is this a server misconfig when the relpath is going to be entered through the user's request parameter? This feels like a malformed request (and so, probably, 400)?

mpacer: I still will it to be less informative of a response and add logging on our side.

chyzzqo2: You are right. For this one the 400 and error message are appropriate.

bookstore/clone.py
mpacer: We have different responses for a file not existing and a file being outside of the notebook basedir on the main notebook server. I'm going to model the response off of the response for a file being outside of the notebook basedir.

https://github.com/jupyter/notebook/blob/e498de6775b3de57f5ff827e562c179b17893064/notebook/services/contents/fileio.py#L262-L263
```python
        if not (os.path.abspath(os_path) + os.path.sep).startswith(root):
            raise HTTPError(404, "%s is outside root contents directory" % path)
```

chyzzqo2: Ok, if that's something you are already doing then it's fine i guess.

mpacer: I still switched it to 404 and added the logging.

bookstore/clone.py
```diff
@@ -300,3 +301,38 @@ def build_post_response_model(self, model, obj, s3_bucket, s3_object_key):
...
+def validate_relpath(relpath, settings, log):
+    """Validates that a relative path appropriately resolves given bookstore settings.
+
+    Parameters
+    ----------
+    model : dict
+        Jupyter Contents API model
+    obj : dict
+        Log (usually from the NotebookApp) for logging endpoint changes.
+    s3_bucket : str
```
chyzzqo2: Think this needs updating

mpacer: oh oops!

mpacer: Fixed!

mpacer:
@chyzzqo2 I think this addresses all your comments now.

PR #158, opened 2019-08-02
Fs cloning handler

mpacer:
This adds a GET handler for file-system cloning.

Requires #157.

PR #159, opened 2019-08-02
End to End testing Cloning Landing page

mpacer:
Builds on #158, adding the route to the server so this can be tested directly.


mpacer:
Spoke with @chyzzqo2 offline and talked through the fact that we should not be adding the route in the first place if the basedir is not correctly configured. 

This required adding stronger checks to the `validate_bookstore` function as well as logging. 

I created #160 to track progress around adding logging to the other bookstore setting values.

willingc:
Thanks @mpacer. :sunny:

PR #163, opened 2019-08-05
Fs cloning api handler

mpacer:
This adds the handler needed to get file system cloning to work. 

I will create a separate PR just for adding the handler to the routes and testing the routes. 

**NB**: This is a bit of a bigger PR, but I didn't want to submit a PR for functionality without tests for that functionality.

### DIFF
--- a/bookstore/clone.py
+++ b/bookstore/clone.py
@@ -409,17 +409,17 @@ class BookstoreFSCloneHandler(IPythonHandler):
 
 
 class BookstoreFSCloneAPIHandler(APIHandler):
-    """Handle notebook clone from a local file system.
+    """Handle notebook clone from an accessible file system (local or cloud).
 
     Provides API handling for ``POST`` and clones a notebook
-    from the local file system.
+    from the specified file system (local or cloud).
 
     Methods
     -------
     initialize(self)
         Helper to access bookstore settings.
     post(self)
-        Clone a notebook from the location specified by the payload.
+        Clone a notebook from the filesystem location specified by the payload.
     build_content_model(self, content, path)
         Helper for creating a Jupyter ContentsAPI compatible model.
 
@@ -433,12 +433,12 @@ class BookstoreFSCloneAPIHandler(APIHandler):
         self.bookstore_settings = BookstoreSettings(config=self.config)
 
     def _get_content(self, path):
-        """Helper for getting content from a local filepath.
+        """Helper for getting content from a specified filepath.
         
         Parameters
         ----------
         path : str
-            Source path from which to get a notebook
+           File system path to a notebook
         """
         self.log.info(f"Reading content from {path}")
         if os.path.splitext(path)[1] in [".ipynb", ".jpynb"]:
--- a/bookstore/clone.py
+++ b/bookstore/clone.py
@@ -406,3 +406,101 @@ class BookstoreFSCloneHandler(IPythonHandler):
     def get_template(self, name):
         """Loads a Jinja template by name."""
         return BOOKSTORE_FILE_LOADER.load(self.settings['jinja2_env'], name)
+
+
+class BookstoreFSCloneAPIHandler(APIHandler):
+    """Handle notebook clone from a local file system.
+
+    Provides API handling for ``POST`` and clones a notebook
+    from the local file system.
+
+    Methods
+    -------
+    initialize(self)
+        Helper to access bookstore settings.
+    post(self)
+        Clone a notebook from the location specified by the payload.
+    build_content_model(self, content, path)
+        Helper for creating a Jupyter ContentsAPI compatible model.
+
+    See also
+    --------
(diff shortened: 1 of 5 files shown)


bookstore/clone.py
```diff
@@ -406,3 +406,101 @@ def construct_template_params(self, relpath, fs_clonepath):
     def get_template(self, name):
         """Loads a Jinja template by name."""
         return BOOKSTORE_FILE_LOADER.load(self.settings['jinja2_env'], name)
+
+
+class BookstoreFSCloneAPIHandler(APIHandler):
+    """Handle notebook clone from a local file system.
```
willingc:
```suggestion
    """Handle notebook clone from an accessible file system (local or cloud).
```

bookstore/clone.py
```diff
@@ -406,3 +406,101 @@ def construct_template_params(self, relpath, fs_clonepath):
...
+
+    Provides API handling for ``POST`` and clones a notebook
+    from the local file system.
```
willingc:
```suggestion
    from the specified file system (local or cloud).
```

bookstore/clone.py
```diff
@@ -406,3 +406,101 @@ def construct_template_params(self, relpath, fs_clonepath):
...
+
+    Methods
+    -------
+    initialize(self)
+        Helper to access bookstore settings.
+    post(self)
+        Clone a notebook from the location specified by the payload.
```
willingc:
```suggestion
        Clone a notebook from the filesystem location specified by the payload.
```

bookstore/clone.py
```diff
@@ -406,3 +406,101 @@ def construct_template_params(self, relpath, fs_clonepath):
...
+    --------
+    `Jupyter Notebook reference on Custom Handlers <https://jupyter-notebook.readthedocs.io/en/stable/extending/handlers.html#registering-custom-handlers>`_
+    """
+
+    def initialize(self):
+        """Helper to retrieve bookstore setting for the session."""
+        self.bookstore_settings = BookstoreSettings(config=self.config)
+
+    def _get_content(self, path):
+        """Helper for getting content from a local filepath.
```
willingc:
```suggestion
        """Helper for getting content from a specified filepath.
```

bookstore/clone.py
```diff
@@ -406,3 +406,101 @@ def construct_template_params(self, relpath, fs_clonepath):
...
+        
+        Parameters
+        ----------
+        path : str
+            Source path from which to get a notebook
```
willingc:
```suggestion
           File system path to a notebook
```

bookstore/clone.py
```diff
@@ -406,3 +406,101 @@ def construct_template_params(self, relpath, fs_clonepath):
...
+    async def post(self):
+        """POST /api/bookstore/fs-clone
+
+        Clone a notebook to the path specified in the payload.
+
+        The payload type for the request should be::
+
+            {
+            "relpath": string,
+            "target_path"?: string
```
willingc:
```suggestion
            "target_path": string
```

willingc: Typo?

bookstore/clone.py
```diff
@@ -270,7 +270,7 @@ def build_content_model(self, content, target_path):
         dict
             Jupyter Contents API compatible model
         """
-        path = self.contents_manager.increment_filename(target_path)
+        path = self.contents_manager.increment_filename(target_path, insert='-')
```
mpacer: To bring @chyzzqo2's comment & response over here:

I added this to ensure consistency between the incrementing patterns for s3 and fs cloning.

bookstore/clone.py
```diff
@@ -406,3 +406,101 @@ def construct_template_params(self, relpath, fs_clonepath):
...
+        ----------
+        content : dict or string
+            dict or string encoded file content
+        target_path : str
+            The the path we wish to clone to, may be incremented if already present.
+
+        Returns
+        --------
+        dict
+            Jupyter Contents API compatible model
```
willingc: Link to Contents API https://jupyter-notebook.readthedocs.io/en/stable/extending/contents.html

bookstore/clone.py
mpacer: It's an optional parameter and I was using the typescript convention of adding a `?` after a key for describing the payload type.

willingc: I would leave off the ? and indicate that target_path is optional. As a Python file, I wouldn't expect the typescript convention to be used.

bookstore/clone.py
```diff
@@ -406,3 +406,101 @@ def construct_template_params(self, relpath, fs_clonepath):
...
+    async def post(self):
+        """POST /api/bookstore/fs-clone
+
+        Clone a notebook to the path specified in the payload.
+
+        The payload type for the request should be::
+
+            {
+            "relpath": string,
+            "target_path": string #optional
```
mpacer: Does this successfully capture the idea in a way that would make sense for a python developer?

mpacer: @willingc 
Does this successfully capture the idea in a way that would make sense from a python perspective?

PR #165, opened 2019-08-05
Add fsclone integration test & format improvements

mpacer:
This adds integration tests for file system cloning and is based on #162 and #164.

It also required addressing some minor inconsistencies I was discovering when I attempted to try our integration tests with a genuinely empty notebook.

These have to do with how the notebook stores cell `source` values in memory vs. on disk. Because we were using `json.dumps()` to get our to-disk content, we were not always writing out the appropriate information that was being passed through. This was most obvious in the case of an empty source `[]` which was being written to s3 as `''`. 
Using `nbformat.writes` and `nbformat.from_dict` alleviates the issue.

### DIFF
--- a/bookstore/archive.py
+++ b/bookstore/archive.py
@@ -6,6 +6,7 @@ from typing import Dict
 from typing import NamedTuple
 
 import aiobotocore
+import nbformat
 from notebook.services.contents.filemanager import FileContentsManager
 from tornado import ioloop
 
@@ -151,7 +152,7 @@ class BookstoreContentsArchiver(FileContentsManager):
             )
             return
 
-        content = json.dumps(model["content"])
+        content = nbformat.writes(nbformat.from_dict(model["content"]))
 
         loop = ioloop.IOLoop.current()
 
--- a/bookstore/tests/__init__.py
+++ b/bookstore/tests/__init__.py
@@ -1,0 +1,3 @@
+import os
+
+test_dir = os.path.realpath(os.path.dirname(__file__))
--- a/bookstore/tests/test_clone.py
+++ b/bookstore/tests/test_clone.py
@@ -28,9 +28,10 @@ from bookstore.clone import (
 )
 from bookstore.utils import TemporaryWorkingDirectory
 
+from . import test_dir
+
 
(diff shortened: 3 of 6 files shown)


bookstore/clone.py
```diff
@@ -270,7 +270,7 @@ def build_content_model(self, content, target_path):
         dict
             Jupyter Contents API compatible model
         """
-        path = self.contents_manager.increment_filename(target_path)
+        path = self.contents_manager.increment_filename(target_path, insert='-')
```
chyzzqo2: Seems unrelated, what was the reason for this?

bookstore/clone.py
```diff
@@ -406,3 +406,101 @@ def construct_template_params(self, relpath, fs_clonepath):
...
+
+        The response payload should match the standard Jupyter contents
+        API POST response.
+        """
+        model = self.get_json_body()
+
+        relpath = model.get("relpath", "")
+        fs_clonepath = validate_relpath(relpath, self.bookstore_settings, self.log)
+
+        target_path = model.get("target_path", "") or os.path.basename(os.path.relpath(relpath))
```
chyzzqo2: If we are going to except a target path, we also need to make sure it is inside of the servers notebook directory. Or is that validation already done somewhere else?

bookstore/clone.py
mpacer: To create consistency between the S3 and FS clone APIs file name conflict policies. I can yank this out and make a separate PR if you'd like.

chyzzqo2: No it's fine to leave in

bookstore/clone.py
mpacer: `self.contents_manager.save(content_model, content_model['path'])` has functionality that strips any prefixed `/` (making it effectively a relpath) and then requests to save it relative to the notebook's base directory.

That in turn runs `_get_os_path` on from the `filio.FileManagerMixin` to make sure that once it's resolved to the real path it is still kept underneath the basedir https://github.com/jupyter/notebook/blob/e498de6775b3de57f5ff827e562c179b17893064/notebook/services/contents/fileio.py#L243-L264

PR #171, opened 2019-08-16
Improve button

### LINKED ISSUE #170: Disable clone button after click in clone.html
Currently when you click the clone button, nothing happens until we get the response from bookstore that we use to redirect to the notebook UI.

This means that someone might click the button a bunch of times and end up with (for example) 16 copies of the same notebook.

We should probably be able to fix this by disabling the button on click so only one request is issued until we get a response.

mpacer:
Closes #170.

This disables the button when clicked and has some style improvements when hovering over the button.


bookstore/clone.html
```diff
@@ -65,7 +70,8 @@
 
         var button = document.getElementById("mybutton");
         button.onclick = function(event) {
-          var post_json = {{ post_model | tojson(indent=2)}}
+          button.disabled=true;
```
stormpython: minor nit pick - create space around the `=` sign.

mpacer: fixed!

mpacer:
> Do we need to re-enable the button after the post response? I could see a UI where the clone action doesn't leave the page and you wouldn't be able to re-initiate the clone action. If this isn't a concern at this time the change looks good to me.

Right now the post request is being populated statically from the jinja template and so I am not sure in what situation the first post request would fail when an second (or otherwise additional) post request would succeed.

MSeal:
I was thinking of any situation in which the button is reused, or the redirect happens in a new tab. Then your button would be permanently disabled on the original tab. Usually js buttons are implemented with a callback so that when the post response returns it re-enables the button. If you redirect to a new page it's fine, but if the implementation changes or gets reused it will also behave as expected. It was mostly a question for if any non-redirect pattern was expected. If so it should have an reenabled callback imo. If it's not a concern it's not blocking a merge.

PR #175, opened 2019-11-25
Use paths not fully qualified domain names within bookstore

mpacer:
Currently by creating the entire the entire domain name, our way of constructing urls is not compatible with a bunch of proxy setups. 

Because we're not changing domain/origin with our links we can just use paths to define our redirect location and our api endpoint for the POST.

This switches us to use that logic.

As I went through making that change I discovered that our tests were not actually testing appropriately for the way that `base_url` and `default_url` on the Jupyter server work. 

Because our tests would break when I switched us to using paths, I added the logic needed to match the behaviour expected by the Jupyter server & the Tornado application. 

This primarily consisted of removing any cases where `self.base_url` did not begin and end with "/".  That required
- adding a default `base_url="/"` to all of our test config (to match the code in [notebook/notebookapp.py#L1048-L1062](
https://github.com/jupyter/notebook/blob/f185f4127d0cdfcca501b8c10983c2b23035c3c3/notebook/notebookapp.py#L1048-L1062)
- Dropping the cases of setting `base_url` that did not begin and end with "/".

### DIFF
--- a/bookstore/tests/test_clone.py
+++ b/bookstore/tests/test_clone.py
@@ -65,7 +65,11 @@ class TestCloneHandler(AsyncTestCase):
     def setUp(self):
         super().setUp()
         self.mock_application = Mock(
-            spec=Application, ui_methods={}, ui_modules={}, settings={'jinja2_env': Environment()}
+            spec=Application,
+            ui_methods={},
+            ui_modules={},
+            # base_url in Tornado settings must begin and end with '/'
+            settings={'jinja2_env': Environment(), "base_url": "/"},
         )
 
     def get_handler(self, uri, app=None):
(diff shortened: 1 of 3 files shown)


bookstore/tests/test_clone.py
```diff
@@ -119,28 +123,26 @@ def test_gen_template_params(self):
...
-            )
-            output = success_handler.construct_template_params(
-                s3_bucket="hello", s3_object_key="my_key"
-            )
-            assert expected == output
+        mock_app = Mock(
+            spec=Application,
+            ui_methods={},
+            ui_modules={},
+            settings={'jinja2_env': Environment(), "base_url": "/my_base_url/"},
```
rgbkrk: Nice, glad you have an explicit check with a `/` for `base_url` and a different one. 😄

Issue #172, opened 2019-09-11
clone a specific version notebook 


rakeshpatri:
I dont find a way to specify versionId from s3 to the clone API, so that i can fetch a specific versioned notebook.
Let me know if there is way to do this in current implementation.

mpacer:
There is no way currently but it is planned to support this.

One of the barriers has been in testing this. We use minio for our integration test, and it does not support versioned buckets and so we may need to either only unit test this functionality or switch to actually using S3 for our integration tests (which introduces its own set of complexities).

mpacer:
@rakeshpatri This is now supported as of 2.5.1, see #176 for the implementation details and https://bookstore.readthedocs.io/en/latest/changelog.html#id3 (version 2.5.1 for the details on how to use it). 

Feel free to comment on the issue or create a new one if you have remaining issues, but I'm going to close this since it has been addressed.

PR #176, opened 2019-12-02
S3 versioned cloning

mpacer:
This adds the ability to clone an old version from a versioned s3 bucket. 

Addresses #172 

Unfortunately we cannot test for this functionality in our minio based integration tests. This is because minio does not support versioned buckets:
see https://github.com/minio/minio/pull/6494 for the latest efforts to implement ending in a decision to not do so and https://github.com/minio/minio/issues/1309 for some of the original discussion around it.

This pulls out functionality around creating the kwargs to send to S3 so that we can test that logic independently.

This also updates the API docs.

@MSeal @captainsafia I added you as reviewers per your request earlier today at the team meeting.

### DIFF
--- a/bookstore/clone.py
+++ b/bookstore/clone.py
@@ -82,7 +82,7 @@ class BookstoreCloneHandler(IPythonHandler):
         Helper to access bookstore settings.
     get(self)
         Checks for valid storage settings and render a UI for clone options.
-    construct_template_params(self, s3_bucket, s3_object_key)
+    construct_template_params(self, s3_bucket, s3_object_key, s3_version_id=None)
         Helper to populate Jinja template for cloning option page.
     get_template(self, name)
         Loads a Jinja template and its related settings.
@@ -115,13 +115,15 @@ class BookstoreCloneHandler(IPythonHandler):
         if s3_object_key == '' or s3_object_key == '/':
             raise web.HTTPError(400, "Requires an S3 object key in order to clone")
 
+        s3_version_id = self.get_argument("s3_version_id", default=None)
+
         self.log.info(f"Setting up cloning landing page for {s3_object_key}")
 
-        template_params = self.construct_template_params(s3_bucket, s3_object_key)
+        template_params = self.construct_template_params(s3_bucket, s3_object_key, s3_version_id)
         self.set_header('Content-Type', 'text/html')
         self.write(self.render_template('clone.html', **template_params))
 
-    def construct_template_params(self, s3_bucket, s3_object_key):
+    def construct_template_params(self, s3_bucket, s3_object_key, s3_version_id=None):
         """Helper that takes valid S3 parameters and populates UI template
         
         Returns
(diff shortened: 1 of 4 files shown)


docs/source/bookstore_api.yaml
```diff
@@ -9,7 +9,7 @@ info:
   license:
     name: BSD 3-clause
     url: https://github.com/nteract/bookstore/blob/master/LICENSE 
-  version: 2.3.0.dev0
+  version: 2.5.1
```
captainsafia: Is this the correct version bump for this?

mpacer: I was bumping it to match what the version of bookstore itself would be upon release, minus the dev0. 

We don't have a separate versioning scheme for the API (we can create one if we think that's best). I removed the dev0 simply because that when we release this, it would match the published version of the package. 

I'm open to other suggestions as to how to handle this, though.

bookstore/clone.py
```diff
@@ -137,12 +139,12 @@ def construct_template_params(self, s3_bucket, s3_object_key):
...
             redirect_contents_url = url_path_join(self.base_url, self.default_url)
         clone_api_url = url_path_join(self.base_url, "/api/bookstore/clone")
-        model = {"s3_bucket": s3_bucket, "s3_key": s3_object_key}
+        model = {"s3_bucket": s3_bucket, "s3_key": s3_object_key, "s3_version_id": s3_version_id}
         template_params = {
             "post_model": model,
             "clone_api_url": clone_api_url,
             "redirect_contents_url": redirect_contents_url,
-            "source_description": f"'{s3_object_key}' from the s3 bucket '{s3_bucket}'",
+            "source_description": f"'{s3_object_key}'{' version: '+s3_version_id if s3_version_id else ''} from the s3 bucket '{s3_bucket}'",
```
MSeal: Might be better to move the conditional version text into a variable above to make it more readable

docs/source/bookstore_api.yaml
MSeal: Might be good to have the version kept in sync between the two files programmatically when running the `bdist_wheel`. If that's a pain to do, then we should instead just add it explicitly to the RELEASING.md instructions imo.

bookstore/clone.py
mpacer: Sounds good!

docs/source/bookstore_api.yaml
mpacer: I think we could set up bumpversion do handle that, but I think for now the low overhead way of making it part of the RELEASING.md makes sense. I think there are a couple of other places where that would help too.

mpacer:
@MSeal @captainsafia I just rebased this PR on master and pushed updates to address the version concern & the string legibility concern. 

I created #177 so a bumpversion setup can be pursued in a separate PR.

bookstore/clone.py
mpacer: this change has been made in the latest version of this PR

docs/source/bookstore_api.yaml
rgbkrk: I realize I'm coming in with a fresh take on this... The version in this YAML file could be different from the library itself since it's the REST API. 🤷‍♂

rgbkrk:
Since it looks like feedback is addressed and there's another thread for the one point of contention, I'll go ahead and merge this.

PR #182, opened 2021-08-26
Notebook servers include a sock attribute

chyzzqo2:


chyzzqo2:
Integration tests seems broken. Likely the same on master. Anyone with more insight into this setup able to take a look?

